PR #47, opened 2026-03-27
Photon Support

ccev:
hi. this is adding Photon support. Photon is an alternative geocoding service to nominatim; https://github.com/komoot/photon - there's a public url to try it out on at http://photon.komoot.io/. Its main advantage is better address search for use in forward geocoding and other services, like diadem.

It uses https://github.com/OpenCageData data to format addresses based on international formats. 

Additionally, there's a separate compact mode that follows the format of my [poracle-middleman project](https://github.com/ccev/poracle-middleman/). Addresses may look like this `{suburb}: {street}, {city}`, depending on what data is available.

It would probably be possible to expand address format support to nominatim, too.

I made sure the formats look good with a couple different addresses, I will switch to this branch in prod shortly.

Disclaimer that this was made largely by AI. Opus 4.6 and then GPT 5.3 Codex, then reviewed by Opus 4.6 and Gemini 3.1 Pro Preview. I checked results manually.


jfberry (commented):
## Review: Photon Support

Thanks for the contribution! The architecture is clean and the Photon provider integrates well with the existing geocoding framework. A few changes requested before merge:

### 1. Remove `photon_address_format` — use existing `address_format` instead

**Problem:** Adding `photon_address_format` creates configuration sprawl. Users would need to understand two different address format config fields.

**How it currently works:** The geocoder always applies `FormatAddress(config.AddressFormat, addr)` to build the `{{addr}}` DTS field (see `geocoder.go` line ~172). This template has access to ALL address fields including `{{{formattedAddress}}}`. So the user already has full control.

**Requested change:**
- Remove the `photon_address_format` config field from `config.go`, `config.example.toml`, and `geocoding.go`
- Remove the `useCompact` parameter from `NewPhoton()` and the compact format code path
- Have Photon ALWAYS populate `FormattedAddress` using OpenCage templates (the better default)
- Users who want compact format can achieve it through the existing `address_format`:
  ```toml
  address_format = "{{{suburb}}}: {{{streetName}}}, {{{city}}}"
  ```
- Users who want the full OpenCage-formatted address use:
  ```toml
  address_format = "{{{formattedAddress}}}"
  ```
- The `FormatCompactAddress` function can be removed entirely — it's just a specific template that users can express themselves

### 2. Remove personal dev files from `.gitignore`

`shell.nix` and `.idea/*` are personal dev environment files. These should go in your `~/.gitignore_global`, not the project's `.gitignore`. Please revert this change.

### 3. Remove whitespace-only changes in `enrichment.go`

The alignment reformatting of the `Enricher` struct fields adds noise to the diff without functional change. Please revert these whitespace changes — only keep the `county` field addition and the `addGeoResult` change.

### 4. Add tests for the Photon provider and OpenCage formatter

The `ocfmt` package and Photon provider have no tests. Please add at minimum:

**`processor/internal/geocoding/ocfmt/ocfmt_test.go`:**
- Test formatting a US address (has specific US template with state abbreviation)
- Test formatting a DE address (different format: street + number order reversed)
- Test formatting with missing fields (e.g. no street, no postcode)
- Test unknown country code falls back to generic template
- Test that the `worldwide.yaml` parses without error

**`processor/internal/geocoding/photon_test.go`:**
- Test `Reverse()` with a mock HTTP server returning a sample Photon GeoJSON response — verify all Address fields are populated correctly
- Test `Reverse()` with empty features array returns error
- Test `Forward()` with a mock HTTP server — verify results
- Test `photonComponents()` mapping from Photon property names to OpenCage component names
- Test that `FormatAddress` template works with Photon-populated addresses (streetName, city, suburb all present)

### 5. Drain response body before close

In `photon.go` `Reverse()` and `Forward()`, the response body should be drained before close for HTTP/1.1 keep-alive to work correctly:

```go
defer func() {
    io.Copy(io.Discard, resp.Body)
    resp.Body.Close()
}()
```

This matches the pattern we use in the Nominatim and Google providers.

### 6. The `county` field addition is good

Adding `County` to the `Address` struct, `FormatAddress`, `EscapeAddress`, and `addGeoResult` is a clean addition that benefits all providers. Keep this.

---

### Summary

The core Photon provider code and OpenCage formatting are well-implemented. The main change is consolidating address formatting to use the existing `address_format` config instead of adding a new `photon_address_format` field. This gives users more flexibility with less configuration. Tests are needed for confidence in the OpenCage template engine and Photon response parsing.

jfberry:
Hi @ccev — I've opened #75 which cherry-picks your commits onto the `goodbye-node` branch (where they'll land post-merge), addresses the review items from this PR, and extends the OpenCage formatting to Nominatim too. Your commits are preserved with original authorship. Planning to merge #75 into `goodbye-node` and close this one once it's merged. Thanks again for the contribution!

PR #53, opened 2026-03-29
Fix `channel remove` failing when deleteHuman is missing on Discord query controller

unseenmagik:
**Summary**
This PR fixes the Discord `channel remove` command by adding the missing `deleteHuman()` helper to the active query controller.

**Problem**
Running `channel remove` was throwing:

TypeError: client.query.deleteHuman is not a function

The Discord runtime uses `alerter/src/controllers/query.js` for `client.query`, but that controller did not expose `deleteHuman()`, even though several command paths were already calling it.

**Root Cause**
There was a mismatch between query controller implementations:

- `alerter/src/controllers/controller.js `already had deleteHuman()
- `alerter/src/controllers/query.js` did not
- Discord commands such as `channel remove` call `client.query.deleteHuman(...)`

That caused runtime failures when removing registered channels or webhooks.

**Changes**
Added `deleteHuman(whereClause)` to `alerter/src/controllers/query.js`.

The helper:

- finds matching human records
- deletes related tracking rows from dependent tables
- deletes the matching `humans` record(s)
This mirrors the existing cleanup behavior already present in the other controller.

**Result**
`channel remove` can now remove registered channels without crashing, and other command paths that already depend on `client.query.deleteHuman(...)` will also work correctly.

### DIFF
--- a/alerter/src/lib/discord/commando/commands/autocreate.js
+++ b/alerter/src/lib/discord/commando/commands/autocreate.js
@@ -1,4 +1,4 @@
-const { PermissionFlagsBits } = require('discord.js')
+const { ChannelType, PermissionFlagsBits } = require('discord.js')
 const PoracleDiscordMessage = require('../../poracleDiscordMessage')
 const PoracleDiscordState = require('../../poracleDiscordState')
 const { loadConfigJson } = require('../../../configResolver')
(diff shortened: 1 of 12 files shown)


unseenmagik:
`Query` in query.js is a standalone lower-level DB helper, and `Controller` in controller.js is a separate higher-level class. `Query` does not extend `Controller`, and it does not have a controller instance to call into, so making `query.deleteHuman()` call `controller.deleteHuman()` would be the wrong dependency direction.

All current calls use `query.deleteHuman(...)`, not `controller.deleteHuman(...)`.

So we could;

- Keep `deleteHuman` in `Query` as the canonical implementation, and remove the duplicate from `Controller`
- Or we avoid duplication with putting shared logic into a small helper like `controllers/common/deleteHuman.js`, then have both methods call that helper.

PR #75, opened 2026-04-18
Photon geocoder support + raymond-based address_format

jfberry:
Takes over #47 (thanks @ccev) and lands it on `goodbye-node` instead of `main` so it doesn't need a second rebase once Goodbye Node merges. Addresses the review on that PR and extends the OpenCage treatment to Nominatim.

## What's in here

**From #47 (ccev's work, cherry-picked with author preserved):**
- Photon reverse-geocoding provider (`processor/internal/geocoding/photon.go`)
- OpenCage address formatting engine (`processor/internal/geocoding/ocfmt/`) with vendored `worldwide.yaml` (~4000 lines of country-specific templates)
- New `county` component field on the `Address` struct
- Performance / validation improvements from ccev's follow-up commits

**Review items applied (see comments on #47):**
1. Dropped `photon_address_format` config + `FormatCompactAddress` helper. FormattedAddress is now always populated via OpenCage; users drive any other shape via the standard `address_format` template.
2. `.gitignore` editor-specific entries reverted (already covered by the branch's `.idea/` / `*.iml` rules).
3. Whitespace-only reformat of `enrichment.Enricher` reverted; only the substantive `county` + `addGeoResult` additions kept.
4. Tests added: `ocfmt_test.go` (US / DE / missing fields / unknown country / YAML parse), expanded `photon_test.go` with `httptest`-backed Reverse + Forward + empty-features and 500 error paths.
5. Response-body drain added on Reverse/Forward error paths so HTTP/1.1 keep-alive works — matches the nominatim/google pattern.
6. `county` field kept.

**Additional: unify address handling across providers.**
- **Nominatim now routes through ocfmt** the same way Photon does, so FormattedAddress is country-idiomatic regardless of provider. Nominatim's raw `display_name` is preserved as a new `Address.DisplayName` field, accessible via `{{{displayName}}}` for operators who relied on the old shape.
- **`address_format` compiles through raymond** (our Handlebars fork — already in the binary for DTS). PoracleJS templates using `{{#if suburb}}{{suburb}}: {{/if}}…` start working again, and the `FormatCompactAddress` helper we dropped is trivially rebuildable in template form.

Compiled once at Geocoder construction; executed per address — no per-call parse cost. The legacy `{{{field}}}` templates continue to work unchanged (triple-brace is unescaped in both the old replacer and Handlebars).

## ⚠️ Default change — worth a second pair of eyes

The in-code default for `locale.address_format` was `"{{{streetName}}} {{streetNumber}}"`; this PR flips it to `""` (empty). Blank means "skip raymond entirely and use the OpenCage-formatted `FormattedAddress` directly."

- **Fresh installs / installs that never touched `address_format`:** `{{addr}}` now renders as the country-idiomatic full address (e.g. `"Unter den Linden 1, 10117 Berlin, Germany"`) instead of just the street line (`"Unter den Linden 1"`). That's a visible output change in every DTS template that uses `{{addr}}` / `{{{addr}}}`.
- **Installs that set `address_format` explicitly in their `config.toml`:** unaffected — their template wins.
- **Installs that deliberately relied on the old default without overriding:** behaviour changes. Restoring the previous output is one config line:
  ```toml
  [locale]
  address_format = "{{{streetName}}} {{streetNumber}}"
  ```

**Arguments for flipping the default:**
- The OpenCage string is almost always more useful than just the street line — it's what PoracleJS users typically hand-templated to approximate.
- With ocfmt wired in, the country-idiomatic default is genuinely good (tested across US, DE, FR, GB, JP).
- Keeps the zero-config "just works" story coherent: new operators don't have to learn the template syntax to get a sensible `{{addr}}`.

**Arguments against:**
- It's a behavioural break for existing installs who used the compiled-in default. Their alerts' `{{addr}}` text will change on upgrade without any config edit on their part.
- Most operators in practice have a custom `address_format` already (this is a commonly-edited line), so the blast radius may be small — but I don't have hard numbers on that.

**Alternatives to consider:**
1. **Keep the old default** (`"{{{streetName}}} {{streetNumber}}"`) to preserve upgrade semantics, and just document the empty-string option.
2. **Flip the default** (current PR behaviour) and add a CHANGELOG / release-note entry calling it out.
3. **Compromise:** keep the old default in-code for existing installs, add a comment in `config.example.toml` recommending blank for new ones.

I'm leaning 2 but happy to revert to 1 if the upgrade-break concern outweighs the better default. Flagging explicitly so it's a deliberate choice, not a drive-by.

## FormatCompactAddress equivalent

Example shipped in `config/config.example.toml`:

```toml
address_format = "{{#if suburb}}{{suburb}}{{else}}{{city}}{{/if}}:{{#if streetName}} {{streetName}}{{#if streetNumber}} {{streetNumber}}{{/if}}{{/if}}{{#if suburb}}, {{city}}{{/if}}"
```

Produces:
- `"Mitte: Friedrichstrasse 42, Berlin"` when a suburb is known
- `"Springfield: Main Street 1"` when only city is
- `"Munich:"` when only city is known and no street

Verified by `TestAddressTemplateCompactEquivalent` in `processor/internal/geocoding/address_test.go`.

## Closes
- #47 — this replaces it. Supersedes, doesn't discard: every ccev commit is preserved with original authorship.

### DIFF
--- a/config/config.example.toml
+++ b/config/config.example.toml
@@ -113,7 +113,29 @@ strict_locations = false
 timeformat = "en-gb"
 time = "LTS"
 date = "L"
-# address_format - how an address is constructed into the {{addr}} tag
+# address_format - Handlebars template used to build the {{addr}} tag from
+# the reverse-geocode result. Compiled once at startup; rendered per address.
+# Available fields (all providers):
+#   formattedAddress  - country-idiomatic string from OpenCage templates
+#   displayName       - the provider's own formatted string (Nominatim only)
+#   streetName, streetNumber, suburb, neighbourhood, city, town, village,
+#   county, state, zipcode, country, countryCode, flag
+#
+# Leaving this empty (or using "{{{formattedAddress}}}") renders the OpenCage
+# country-idiomatic default, which is usually what you want.
+#
+# Examples:
+#   address_format = "{{{streetName}}} {{streetNumber}}"
+#       Just the street line (current default for new installs).
+#   address_format = "{{{formattedAddress}}}"
+#       Full country-idiomatic address (e.g. "Unter den Linden 1, 10117 Berlin, Germany").
+#   address_format = "{{{displayName}}}"
+#       Nominatim's original display_name layout, for operators migrating from
+#       older PoracleJS installs who relied on that shape.
+#   address_format = "{{#if suburb}}{{suburb}}{{else}}{{city}}{{/if}}:{{#if streetName}} {{streetName}}{{#if streetNumber}} {{streetNumber}}{{/if}}{{/if}}{{#if suburb}}, {{city}}{{/if}}"
+#       Equivalent to the old Photon/PoracleJS compact format —
+#       "Mitte: Friedrichstrasse 42, Berlin" when suburb is present,
+#       "Springfield: Main Street 1" when only city is known.
 address_format = "{{{streetName}}} {{streetNumber}}"
 # language - secondary language used for 'alt' translation in DTS; useful for
 #   retrieving english language detail for web links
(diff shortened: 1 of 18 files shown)


ccev:
have this running, can confirm photon works and formattedAddress, as well as compactAddress look good.

PR #87, opened 2026-05-10
Quest summary: buffered, scheduled grouped delivery

jfberry:
## Summary
- Adds a per-user, per-alert-type **summary scheduler** that buffers matched quests in memory and dispatches them as grouped messages on a wall-clock schedule (`!summary quest settime mon07:30 weekend10:00`, etc.).
- New `summary` keyword on `!quest` opts a rule in (sets `clean & 4`); matched quests skip immediate delivery and accumulate in `tracker.SummaryBuffer`, snapshotted to `config/.cache/summary-buffer.json` on shutdown.
- New `questSummary` DTS template (with multi-pin static map) renders one message per `(rewardType, reward)` group; oversize groups are chunked across multiple messages with pagination so embeds stay under Discord's 6000-char/25-field limits.
- `!summary` command (`settime`/`cleartime`/`now`) and `/api/summaries/{id}/{alertType}` endpoints; `!summary quest now` flushes the buffer on demand. Dispatches via `Dispatcher.DispatchBypass` so summary sends aren't swallowed by the rate limiter.
- Adds `every` / `everyday` day prefix for settime so users can write `every08:00` instead of listing all seven days; profile + new `summary` help DTS document the full prefix syntax.

## Notable bits
- `db.IsSummary(clean)` bit (clean & 4) is the on-disk signal; matching splits matched users into `immediate` vs `buffered` slices.
- `SummaryScheduler` wakes on the same minute marks as the profile scheduler; reads humans from state, dispatches buffered groups through the render pool.
- Tile mode: `questSummary` follows the standard `/staticmap/poracle-questsummary` URL pattern; per-pin `imgUrl` injected via the existing autoposition helper.
- `tools/webhook-search.py` learned quest-aware result display + pokemon/item name → ID resolution for `--reward`.

## Test plan
- [ ] `cd processor && go test ./... -count=1`
- [ ] `!quest pikachu summary` — rule shows `summary` indicator in `!tracked`
- [ ] `!summary quest` — status (schedule + buffer count)
- [ ] `!summary quest settime every08:00` — round-trips through JSON, displays "every day 08:00"
- [ ] `!summary quest now` — buffered quests dispatched as one or more grouped messages, multi-pin map renders
- [ ] Restart processor mid-buffer — snapshot reload preserves entries
- [ ] Oversize group (>25 fields) chunks into multiple messages, each chunk's map shows only its own pins
- [ ] Rate-limit not triggered by a summary dispatch fanning out to one user


jfberry:
Two consistency gaps surfaced while writing user docs for this PR:

**1. `!help quest` doesn't mention the new `summary` keyword.**

`fallbacks/dts/help/quest.json` isn't touched in this PR, but `summary` is now a valid `!quest` option. Users who do `!help quest` won't discover it. Worth adding a field to the quest help template alongside the existing keyword list, e.g.:

> **Buffered delivery** — add `summary` to defer matched quests into a digest. See `!help summary`.

**2. `!profile` should gain `cleartime` for parity with `!summary`.**

`!summary <type>` exposes `settime` / `cleartime` / `now`. `!profile` exposes `settime` but no `cleartime` — once a user sets an auto-switch schedule there's no documented way to remove it short of recreating the profile (or running `settime` with no args, if that even works). For the day-prefix syntax (`every`, `everyday`, `weekday`, etc.) to feel like a shared mini-language across `!profile` and `!summary`, the schedule lifecycle commands should match: at minimum, add `!profile cleartime` and update `fallbacks/dts/help/profile.json` to document it.

Both items are small but bite the discoverability story: the docs page I just wrote points users at `!help summary` from the quest page, but the reverse path (`!help quest` → mentions `summary`) is silent, and the profile/summary symmetry I'm pitching in `userguide/profiles.md` doesn't quite hold without `cleartime` on profiles.

Happy to update the docs site once these land.

jfberry:
### Code review

Found 3 issues:

1. Buffered quest matches skip `filterValidation` (CLAUDE.md says "Phase 1 — pre-filter (match time): After matching, each webhook handler calls `filterBlocked(matched)` …"; `filterValidation` is the matching external-validator hook). `matched` users get both filters; `buffered` users go straight to `bufferQuestMatches`, so a user denied by the operator's configured `/api/validate` endpoint still gets buffered and delivered at summary time.

https://github.com/jfberry/PoracleNG/blob/12b2d131434c32bdcef25737c5a1da38924c7b29/processor/cmd/processor/quest.go#L70-L80

2. Docs contradict code on `clean` bit propagation. CLAUDE.md says "the source rule's `clean` bit propagates so TTH-based deletion still works" and DTS.md repeats it, but `DispatchQuestSummary` builds the per-user `MatchedUser` with `Clean` defaulting to zero and the inline comment says "summary-bucket clean bit is intentionally not propagated to the summary message." Either the code should propagate the bit (and TTH-clean would work as documented) or both docs should say it doesn't.

https://github.com/jfberry/PoracleNG/blob/12b2d131434c32bdcef25737c5a1da38924c7b29/processor/cmd/processor/quest_summary_dispatch.go#L104-L112

https://github.com/jfberry/PoracleNG/blob/12b2d131434c32bdcef25737c5a1da38924c7b29/CLAUDE.md#L278-L280

3. `tileMaptype` case mismatch silently drops the per-type tileserver override. `quest_summary.go:82` looks up `"questsummary"` (all lowercase); `config.example.toml:406` documents the override key as `questSummary` (camelCase). `StaticMapType` is a case-sensitive `map[string]string` loaded straight from TOML, so an operator's `[geocoding.static_map_type] questSummary = "multiStaticMap"` is silently ignored.

https://github.com/jfberry/PoracleNG/blob/12b2d131434c32bdcef25737c5a1da38924c7b29/processor/internal/dts/quest_summary.go#L80-L90

https://github.com/jfberry/PoracleNG/blob/12b2d131434c32bdcef25737c5a1da38924c7b29/config/config.example.toml#L405-L407

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

jfberry:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance on the commits since the [previous review](https://github.com/jfberry/PoracleNG/pull/87#issuecomment-4431173530) (clean-bit + TTH propagation, separate summary rate-limit bucket, range+step settime parser, status-line + Prometheus gauge, profile cleartime parity, help text updates, go fix modernizations). All three prior issues are properly addressed.

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #88, opened 2026-05-10
Pokemon change notifications (monsterChanged + reply threading)

jfberry:
## Summary

- New `monsterChanged` template type that fires on post-initial encounter changes (species / form / gender / encountered-state / weather-boost). Initial sightings still use `monster`.
- Per-user reply threading: change alerts arrive as a Discord reply / Telegram `reply_to_message_id` pointing at the original `monster` alert. Driven by a new O(1) reply index in `delivery.MessageTracker` keyed on `ReplyKey = encounterID`.
- `EncounterTracker` (`internal/tracker/encounter.go`) detects change types and stores a PVP-stripped copy of the prior webhook so `monsterChanged` templates can reference an `{{original.*}}` namespace mirroring the regular `monster` field set (minus PVP).
- Tracker storage is gated on `len(matched) > 0`: pokemon nobody is tracking never enter the tracker, since they cannot produce a per-user change event later. Dedup remains separate via `ps.duplicates.CheckPokemon`.
- Feature flag: `[tracking] pokemon_change_tracking` (off by default).
- Bundled `monsterChanged` DTS entries for Discord and Telegram in `fallbacks/dts.json`.
- Docs: API.md and DTS.md updated with the `original.*` field set and `monsterChanged` template usage.

## Test plan

- [ ] `cd processor && go build ./cmd/processor` succeeds
- [ ] `cd processor && go test ./...` green (delivery, tracker, dts, cmd/processor suites all touched)
- [ ] With `pokemon_change_tracking = true`, replay two webhooks for the same encounter ID where a user matches both sightings and the second represents a real change — confirm the second arrives as `monsterChanged` and is a reply to the first
- [ ] Same setup but the user only matches the second sighting — confirm they get a plain `monster` (no reply, no `original.*`)
- [ ] With `pokemon_change_tracking = false`, all sightings render as plain `monster` and no reply threading occurs
- [ ] Confirm pokemon with zero matchers no longer touch the encounter tracker (look for tracker map growth under load on a feed with many low-IV spawns)


jfberry:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #95, opened 2026-05-15
pokemon: fire monsterChanged on ChangeEncountered for prior-only users

jfberry:
## Summary

Users with an IV/CP/level filter on their tracking rule saw `monsterNoIv` on the wild webhook (matcher skips encounter-only filters when CP is unknown). When the encounter webhook arrived with revealed stats below their filter, they landed in priorOnly with \`change.Type=ChangeEncountered\` — and the dispatcher silently dropped the bucket. The optimistic T1 alert was left dangling with no follow-up.

## Cause

Earlier I added \`in.change.Type != tracker.ChangeEncountered\` to bucket 2 in \`dispatchPokemonAlert\` on the reasoning that users who still match at T2 get a regular \`monster\` reply via the matched path. That's true for catch-all rules (e.g. \`!track pikachu\`), but for IV/CP-filtered users whose reveal fails the filter, dropping the bucket entirely means they never hear what the IVs actually were.

## Fix

- Drop the \`ChangeEncountered\` skip in bucket 2.
- \`changeTypeBucket\` now maps \`ChangeEncountered\` to \`"stats"\` alongside \`ChangeWeatherBoost\` — same pokemon, different effective stats.
- Replaced \`TestDispatchPokemonAlert_EncounteredChange_PriorOnly_NoJob\` with \`_FiresStats\` (1 RenderJob, IsChange=true, ReplyKey set, Clean inherited, \`changeType="stats"\`).
- Updated DTS.md.

## Test plan

- [x] \`go build ./...\` clean.
- [x] \`go vet ./...\` clean.
- [x] Full test suite green.
- [ ] Manual verify: \`!track pikachu iv90\`, wait for a sub-90% IV Pikachu encounter, confirm \`monsterChanged\` reply now arrives.

### DIFF
--- a/DTS.md
+++ b/DTS.md
@@ -360,12 +360,11 @@ fired the alert. Use these in templates to switch wording or styling:
 | `changeType` | Fires for | Meaning |
 |---|---|---|
 | `species` | `ChangeSpecies`, `ChangeForm`, `ChangeGender` | Identity change. Most commonly community-day re-classifications, or the known "A/B pokemon" server bug where Golbat reports a different species ID for the same encounter. |
-| `stats` | `ChangeWeatherBoost` | Same pokemon, weather boost shifted. Post-boost IVs and CP differ from the originally-alerted state. This is the most common cause in practice. |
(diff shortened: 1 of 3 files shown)


jfberry:
Closing — the wild→encounter scenario this PR addresses doesn't exist in practice. The matcher rejects rules with IV filters when CP=0 (since data.IV=-1 fails MinIV>=0 checks), so users with strict IV filters never see monsterNoIv on the wild webhook. Real fix lands separately to detect raw IV drift between encountered webhooks.

PR #99, opened 2026-05-16
Discord slash commands (optional)

jfberry:
This branch adds a `/command` surface to PoracleNG alongside the existing `!command` text commands, AND a new `!poracle-admin` (alias `!pa`) admin umbrella for live operations. Slash commands are optional, off by default, and additive — your users can keep typing `!track pikachu iv95` and nothing changes for them.

## Why try it

- Discord's native slash UI gives users autocomplete dropdowns for pokemon names, raid bosses, items, areas, profiles, and their own existing tracking rules.
- The bot's reply only appears to the user who ran the command (no channel noise from setup chatter).
- Newcomers don't have to know the text-command grammar — they just see the option list.
- **Operators** get a single `!poracle-admin` umbrella for live ops (slash registration, reload, status, maintenance pause, ratelimit inspection, …) so they don't have to restart the processor to fix common things.

## Turning it on

Add to `config/config.toml`:

```toml
[discord.slash_commands]
enabled = true
register_globally = false
guilds = ["<your test guild ID>"]
```

Restart the processor. You should see startup log lines like:

```
slash: dispatcher enabled, guilds=[123456789012345678], force_sync=false
slash: guild 123456789012345678 sync OK — 23 commands, fingerprint 4f3a1c2e9b8d6f02
```

Type `/` in any channel the bot is in — you'll see the command list pop up.

### Guild vs global registration

| Setting | Behaviour |
| --- | --- |
| `register_globally = false` + `guilds = [...]` | Per-guild registration. Updates appear within seconds. Use this for testing. |
| `register_globally = true` | Single global registration. Updates can take **up to 1 hour** to propagate to clients. Use this once you're happy with the surface. |

**Switching between the two leaves stale registrations behind**, and Discord will show duplicate commands. You can clean up either via the live admin command (no restart) or with the old CLI flags:

```
!poracle-admin slash clear-global       # live, after switching to guild mode
!poracle-admin slash clear-guild 12345  # live, removes one guild's stale set
```

CLI fallback (for when the gateway is broken):
```
./poracle-processor -basedir .. -clear-global-slash-commands
./poracle-processor -basedir .. -clear-guild-slash-commands
```

## Commands available (user surface)

| Read-only | Mutating |
| --- | --- |
| `/version` `/tracked` `/help` `/info` `/language` `/start` `/stop` | `/track` `/raid` `/egg` `/quest` `/invasion` `/incident` `/lure` `/nest` `/maxbattle` `/gym` `/fort` |
[...]

PR #105, opened 2026-05-18
Raid RSVP-update template + egg→raid reply chain

jfberry:
Branched off `slash-commands-design` (PR #99) so the two can be tested together. Two improvements to raid + egg alerts:

## What you get

**1. Optional compact "RSVP Update" template.** When a raid's RSVP counts change, you can have the bot post a short embed (timeslot + going/maybe counts + map links) instead of repeating the full raid card. Off by default — operator opts in by copying a template into `config/dts/`.

**2. Egg → raid → RSVP reply chain.** Discord/Telegram thread the lifecycle: when a tracked egg hatches, the raid notification arrives as a reply to the egg. When RSVPs change, those messages reply to the raid. Whole sequence reads as one thread in the channel. Works automatically, no extra setup. Requires you to track both `!egg` and `!raid` if you want the egg→raid leg.

**Cleanup timing** for the compact RSVP-update messages is per-message — they auto-clean shortly after their latest party time, not at raid end, so they don't linger past the meaningful window.

## Setup

To enable the compact RSVP-update template:

```bash
cp -r examples/dts/rsvpChanges/* config/dts/
!poracle-admin reload dts
```

That's it. Tracking rules with `rsvp` or `rsvp_only` (with `clean` set, so the messages get auto-deleted) will now use the compact card on RSVP changes.

To disable: delete the file from `config/dts/`, reload DTS. Falls back to the full raid card (current behavior).

The egg→raid reply chain needs no setup — it activates as soon as you have both `!egg <level>` and `!raid <level>` rules in place. Edit-mode (`!raid level5 rsvp edit`) is unaffected — that path still edits one message in-place.

## What to look for during testing

`docs/raid-rsvp/SMOKE.md` has a checklist you can walk through against a live Golbat raid stream. The key items:

- Raid alert arrives as a reply under its egg alert (with both tracked).
- With the compact template installed: RSVP changes post compact cards, threaded as replies to the raid card.
- Compact cards disappear shortly after the last party time.
- Edit mode is unchanged.
- Operators who don't copy the example template see no behavior change beyond the egg→raid reply.

## Reporting

Most likely failure modes:
1. Raid not replying to egg — check the egg actually fired (some operators only track raids).
2. Compact card not appearing — check the template file is in `config/dts/` (not `config/dts.json`) and `!pa reload dts` ran without errors.
3. Compact card auto-delete fires at raid end instead of latest timeslot — confirm this branch is deployed (not a stale main).

Include a screenshot of what Discord shows + the `!tracked` output for the relevant rule.


jfberry:
Superseded by the develop-as-early-adopters branch model. The raid-rsvp work is integrated into the develop branch; PR #121 is the new develop -> main release PR. Review history preserved here.

PR #115, opened 2026-05-20
Fix !track when only min is provided

lglerup:
Very basic fix, where maxSize is set to Min, if no range is provided in !track
Also follows the docs already: https://jfberry.github.io/PoracleNG-docs/userguide/pokemon/#track-by-size

Previous behaviour
  - size1 → size=1, maxSize=5 → size: XXS-XXL
  - size3 → size=3, maxSize=5 → size: M-XXL
  - size1-3 → size=1, maxSize=3 → size: XXS-M (explicit range, unchanged)

New behaviour
  - size1 → size=1, maxSize=1 → size: XXS (exact match)
  - size3 → size=3, maxSize=3 → size: M (exact match)
  - size1-3 → size=1, maxSize=3 → size: XXS-M (explicit range, unchanged)


jfberry:
Thanks for noticing, this is indeed a variance to PoracleJs

PR #117, opened 2026-05-21
[codex] Fix gym battle cooldown deduplication

Mygod:
## Summary
- preserve the first gym under-attack alert while suppressing repeated same-state battle webhooks
- move battle cooldown state into the gym-state transition so old state and cooldown are decided under one lock
- align battle-change matching with PoracleJS behavior and cover cold-gym concurrency with targeted tests

## Root Cause
NG started and checked the gym battle cooldown outside the gym-state transition. Under concurrency, duplicate first in-battle webhooks for a cold gym could observe inconsistent cooldown and old-state ordering, allowing duplicate battle alerts.

## Validation
- go test ./internal/tracker ./internal/matching
- go test ./cmd/processor -run '^$'
- go test -race ./internal/tracker
- git diff --check

### DIFF
--- a/processor/cmd/processor/gym.go
+++ b/processor/cmd/processor/gym.go
@@ -55,17 +55,14 @@ func (ps *ProcessorService) ProcessGym(raw json.RawMessage) error {
 		// Resolve in-battle
 		inBattle := bool(gym.IsInBattle) || bool(gym.InBattle)
 
-		// Battle cooldown: during battles, Golbat sends frequent updates.
-		// GymInBattleCooldown reports the pre-existing cooldown state, then
-		// starts/refreshes it for in-battle webhooks. This keeps the first
-		// under-attack update eligible and suppresses repeated same-state ones.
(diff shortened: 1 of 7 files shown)


jfberry (approved):
Verified the cooldown semantics against PoracleJS `src/controllers/gym.js:21` and `src/app.js:749-760` — the matcher change (drop `!OldInBattle`) brings parity with JS (which only checks `data.inBattle`), and folding cooldown+state under one lock closes the cold-gym race. Tests are well-scoped: `TestGymStateBattleCooldown` pins the basic transitions, `TestGymStateBattleCooldownConcurrentColdGym` pins the concurrency invariant, and `TestGymBattleActiveMatchesAfterPriorBattleState` pins the matcher change. Thanks for the careful work — clean removal of the redundant `gymBattle` ttlcache too.

Mygod:
Wow can we have this merging speed for my Golbat PR too? 😄

PR #126, opened 2026-05-24
api: add user language endpoint

ccev:
made by gpt 5.5

### DIFF
--- a/API.md
+++ b/API.md
@@ -450,6 +450,20 @@ Toggle admin disable flag.
 {"state": true}
 ```
 
+### POST /api/humans/{id}/language
+
+Set a user's alert and bot-response language. If `[general.available_languages]` is configured, `language` must be one of those configured language codes; otherwise any non-empty language code is accepted.
+
+```bash
+curl -X POST -H "X-Poracle-Secret: secret" -H "Content-Type: application/json" \
+  http://localhost:3030/api/humans/123456789/language \
+  -d '{"language": "de"}'
+```
+
+```json
+{"status": "ok", "language": "de"}
+```
+
 ### POST /api/humans/{id}/setLocation/{lat}/{lon}
 
 Update user location. Validates against area restrictions if area security is enabled.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -519,6 +519,7 @@ All API endpoints are accessed via the processor (port 3030). The processor hand
 | POST | `/api/humans/{id}/start` | Enable a user |
 | POST | `/api/humans/{id}/stop` | Disable a user |
 | POST | `/api/humans/{id}/adminDisabled` | Toggle admin disable flag |
+| POST | `/api/humans/{id}/language` | Set user locale/language |
 | POST | `/api/humans/{id}/setLocation/{lat}/{lon}` | Update user location |
 | GET | `/api/humans/{id}/checkLocation/{lat}/{lon}` | Check if location is in allowed areas |
 | POST | `/api/humans/{id}/setAreas` | Set user's selected geofence areas |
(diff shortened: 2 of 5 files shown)


jfberry (approved):
## Review

Looks good. Consistent with the rest of the `/api/humans/{id}/...` family and uses the established HumanStore boundary correctly.

**Verdict: approve, with two small follow-up nits worth taking but not blocking.**

### Consistency check (passed)

- ✓ `TrackingDeps`, `trackingJSONError`, `trackingJSONOK`, `reloadState` — same helpers as `HandleAdminDisabled` / `HandleSetLocation` / `HandleSetAreas`.
- ✓ Same handler shape: validate `id` → `Humans.Get` → 404 if nil → bind JSON → validate → mutate → reload.
- ✓ Uses `deps.Humans.SetLanguage(id, language)` — the existing HumanStore method, also called by the `!language` command. No raw SQL outside the store. Matches the boundary documented in CLAUDE.md.
- ✓ Route registration in `main.go` follows the existing pattern.
- ✓ Mock store already supports `SetLanguage` so tests don't need new mock plumbing.
- ✓ Case normalization (input lowercase, output preserves operator's configured case via `matched = code`) matches `!language` command at `bot/commands/language.go:43-48`.
- ✓ Test coverage is sensible: success, code-case-folding (`DE` → `de`), unavailable-language rejection, unrestricted-when-no-config branch, missing user, blank input. Each asserts both response code AND that `reloads` was/wasn't bumped.
- ✓ API.md + CLAUDE.md additions match the established table style.

### Nits (non-blocking)

**1. Code duplication with `!language` command** — the `for code := range available { if strings.ToLower(code) == language { matched = code } }` loop is nearly identical to `bot/commands/language.go:43-48`. Worth extracting to `config.ResolveAvailableLanguage(input string) (matched string, ok bool)` so both call sites share the resolution + case-preserving logic. The `!language` command also matches translated and English language names — when the helper lands, both surfaces can pick up that capability for free. Not for this PR; follow-up.

**2. Error response could surface the available codes** — current 400 body says `"language is not available"` with no list. The `!language` command shows the list when the user types a bad value. The admin API probably uses this from a UI where the list is already known, so likely fine, but a `"available": ["de", "en", "fr"]` field in the error body would be helpful for clients constructing UI dropdowns from the response. Cheap to add:

```go
if matched == "" {
    codes := make([]string, 0, len(deps.Config.General.AvailableLanguages))
    for code := range deps.Config.General.AvailableLanguages {
        codes = append(codes, code)
    }
    sort.Strings(codes)
    c.JSON(http.StatusBadRequest, gin.H{
        "status":    "error",
        "message":   "language is not available",
        "available": codes,
    })
    return
}
```

Mirrors the shape `validateConfigValues` uses elsewhere for "your input wasn't valid; here are the valid options".

### One thing worth knowing (not a change request)

The user-facing `!language` command **sends a confirmation message in the destination language** ("✅ Sprache geändert nach Deutsch"). This API endpoint doesn't (it's admin-driven; no message to the affected user). That asymmetry is correct — the admin endpoint is for operator tooling, not user-facing language switching — but worth being explicit about in the API.md description so editor authors don't expect the side effect:

> `POST /api/humans/{id}/language` sets the user's stored language. Does **not** send a confirmation message to the user. To switch language with a user-facing confirmation, the user should run `!language <code>` themselves.

### Ready

Tests pass against the existing mock store and don't need any other test scaffolding changed. Endpoint slots cleanly into the `/api/humans/{id}/...` group. Approving — nits are improvements, not blockers.

jfberry:
Merged into `develop` (was retargeted from `main`).

For context on the branch model: PoracleNG runs a develop-as-early-adopters channel. New features land on `develop` first, then promote to `main` in batches via the long-lived release PR #121. Operators on `main` are on the conservative track; operators tracking `develop` will see this endpoint immediately after the next deploy.

Two follow-up nits from the review, not blocking the merge but worth picking up either here or as a follow-up PR:

1. **Shared `ResolveAvailableLanguage` helper** — the available-language matching loop is duplicated between this handler and `internal/bot/commands/language.go:43-48`. Extracting to `config.ResolveAvailableLanguage(input string) (matched string, ok bool)` would let both surfaces share validation + the `!language` command's bonus "match by translated name" feature.

2. **Include `available` codes in 400 error body** — current 400 says `"language is not available"` with no list. Adding `"available": [...]` to the response body lets editor UIs build dropdowns from the response.

Happy to take either as a separate PR or leave them for whoever next touches the area. Thanks for the contribution.

PR #127, opened 2026-05-24
geocoding: per-language reverse geocode results via localized cache keys

ccev:
- Extend Provider interface: Reverse(lat, lon, language string)
- Plumb language through Nominatim (accept-language), Photon (lang), Google (language)
- CacheKeyForLanguage isolates localized results in the on-disk cache
- addLocalizedGeocoding clones PerLangEnrichment then merges per-language address fields
- Base enrichment stores geocodeLatitude/geocodeLongitude for later per-lang lookups
- Tests for language forwarding and cache key isolation

made by gpt 5.5, approved by deepseek v4 pro

### DIFF
--- a/API.md
+++ b/API.md
@@ -1430,7 +1430,7 @@ All invasion grunt types.
 
 ## Geocoding
 
-Reverse geocoding for alert templates is locale-aware. The processor performs a base lookup as a fallback, then performs per-language lookups for the matched users' effective languages and stores those address fields in the per-language DTS layer. Existing DTS fields such as `{{addr}}`, `{{city}}`, `{{streetName}}`, and `{{formattedAddress}}` automatically resolve to the user's localized geocoder result when the provider supports it.
+Reverse geocoding for alert templates is locale-aware. The processor performs a base reverse-geocode lookup that populates the default-locale address fields, then performs per-language lookups for matched user languages that shadow those fields in the per-language DTS layer. Existing DTS fields such as `{{addr}}`, `{{city}}`, `{{streetName}}`, and `{{formattedAddress}}` automatically resolve to the user's localized geocoder result when the provider supports it.
 
 Supported reverse-geocode language parameters:
 
--- a/API.md
+++ b/API.md
@@ -1430,6 +1430,18 @@ All invasion grunt types.
 
 ## Geocoding
 
+Reverse geocoding for alert templates is locale-aware. The processor performs a base lookup as a fallback, then performs per-language lookups for the matched users' effective languages and stores those address fields in the per-language DTS layer. Existing DTS fields such as `{{addr}}`, `{{city}}`, `{{streetName}}`, and `{{formattedAddress}}` automatically resolve to the user's localized geocoder result when the provider supports it.
+
+Supported reverse-geocode language parameters:
+
+| Provider | Parameter |
+|----------|-----------|
+| Nominatim | `accept-language` |
+| Photon | `lang` |
+| Google | `language` |
+
+Reverse geocode cache keys include the language code for localized lookups, so cached German and English address results do not overwrite each other.
+
 ### GET /api/geocode/forward?q={query}
 
 Forward geocode a location name to coordinates.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -186,7 +186,8 @@ Enrichment is computed in three layers:
 - **Map URLs**: Google Maps, Apple Maps, Waze, RDM, ReactMap, RocketMad
 - **Icons**: Pokemon/raid/gym/weather icon URLs via uicons (with scheduled index refresh)
 - **Static map**: Tileserver tile URL (pregenerate or query params), with scanner DB stop lookups
-- **Geocoding**: Reverse geocode address via nominatim/photon/google with pogreb on-disk + ttlcache in-memory cache. Alert rendering adds per-language reverse geocoding for each matched user language (`accept-language` for Nominatim, `lang` for Photon, `language` for Google), and localized cache keys include the language code so address fields (`addr`, `city`, `streetName`, etc.) resolve in the user's locale when the provider supports it.
+- **Geocoding**: Reverse geocode address via nominatim/photon/google with pogreb on-disk + ttlcache in-memory cache.
+- **Localized addresses**: Per-language enrichment adds reverse geocoding for each matched user language (`accept-language` for Nominatim, `lang` for Photon, `language` for Google). Localized cache keys include the language code so address fields (`addr`, `city`, `streetName`, etc.) resolve in the user's locale when the provider supports it.
 
 **Per-language enrichment** (computed once per distinct language among matched users):
 - Translated pokemon/form/move/type/weather/team/generation names using pogo-translations identifier keys (`poke_1`, `move_14`, `form_46`, etc.)
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -186,7 +186,7 @@ Enrichment is computed in three layers:
 - **Map URLs**: Google Maps, Apple Maps, Waze, RDM, ReactMap, RocketMad
 - **Icons**: Pokemon/raid/gym/weather icon URLs via uicons (with scheduled index refresh)
 - **Static map**: Tileserver tile URL (pregenerate or query params), with scanner DB stop lookups
-- **Geocoding**: Reverse geocode address via nominatim/google with pogreb on-disk + ttlcache in-memory cache
+- **Geocoding**: Reverse geocode address via nominatim/photon/google with pogreb on-disk + ttlcache in-memory cache. Alert rendering adds per-language reverse geocoding for each matched user language (`accept-language` for Nominatim, `lang` for Photon, `language` for Google), and localized cache keys include the language code so address fields (`addr`, `city`, `streetName`, etc.) resolve in the user's locale when the provider supports it.
 
 **Per-language enrichment** (computed once per distinct language among matched users):
 - Translated pokemon/form/move/type/weather/team/generation names using pogo-translations identifier keys (`poke_1`, `move_14`, `form_46`, etc.)
--- a/processor/cmd/processor/enrich.go
+++ b/processor/cmd/processor/enrich.go
@@ -226,7 +226,7 @@ func (ps *ProcessorService) enrichInvasion(raw json.RawMessage, language string)
 
 	var perLang map[string]any
 	if ps.enricher.GameData != nil && ps.enricher.Translations != nil {
-		perLang = ps.enricher.InvasionTranslate(base, gruntTypeID, inv.Lineup, inv.ShowcaseRankings, language)
+		perLang = ps.enricher.InvasionTranslate(base, inv.Latitude, inv.Longitude, gruntTypeID, inv.Lineup, inv.ShowcaseRankings, language)
 	}
 
 	return &enrichResult{
@@ -248,7 +248,7 @@ func (ps *ProcessorService) enrichLure(raw json.RawMessage, language string) (*e
 
 	var perLang map[string]any
 	if ps.enricher.GameData != nil && ps.enricher.Translations != nil {
-		perLang = ps.enricher.LureTranslate(base, lure.LureID, language)
+		perLang = ps.enricher.LureTranslate(base, &lure, language)
 	}
 
 	return &enrichResult{
@@ -270,7 +270,7 @@ func (ps *ProcessorService) enrichNest(raw json.RawMessage, language string) (*e
 
 	var perLang map[string]any
 	if ps.enricher.GameData != nil && ps.enricher.Translations != nil {
-		perLang = ps.enricher.NestTranslate(base, nest.PokemonID, nest.Form, language)
+		perLang = ps.enricher.NestTranslate(base, &nest, language)
 	}
 
 	return &enrichResult{
@@ -303,7 +303,7 @@ func (ps *ProcessorService) enrichGym(raw json.RawMessage, language string) (*en
 	var perLang map[string]any
 	if ps.enricher.GameData != nil && ps.enricher.Translations != nil {
 		// Editor preview has no team-change context, so suppress previousControl* fields.
-		perLang = ps.enricher.GymTranslate(base, teamID, -1, -1, language)
+		perLang = ps.enricher.GymTranslate(base, gym.Latitude, gym.Longitude, teamID, -1, -1, language)
 	}
 
 	return &enrichResult{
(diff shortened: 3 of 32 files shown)


jfberry (commented):
## Review summary

The cache-key plumbing, provider parameter wiring, and language normalization are solid. The integration point with the enrichment pipeline is in the wrong place architecturally — it should fold into the existing per-language enrichment step instead of adding an out-of-band render-worker step. That single change also removes most of the other concerns in this review as side effects (hot-path geocoding, map cloning, leak into base, redundant lookup when language matches base).

**Verdict: rework before merge.** Keep the provider-side and cache-side work; rework the integration.

---

## Required: align with the existing per-language enrichment pattern

Every per-type handler (`pokemon.go`, `raid.go`, `gym.go`, `invasion.go`, etc.) already builds `perLang` via the canonical pattern, e.g. `cmd/processor/pokemon.go`:

```go
if ps.enricher.GameData != nil && ps.enricher.Translations != nil {
    perLang = make(map[string]map[string]any)
    for _, lang := range distinctLanguages(allRecipients, ps.cfg.General.Locale) {
        perLang[lang] = ps.enricher.PokemonTranslate(baseEnrichment, &pokemon, lang)
    }
}
```

`PokemonTranslate` / `RaidTranslate` / etc. **is** the per-language enrichment site. Localized geocoding belongs there alongside translated pokemon names, types, and weather. The PR instead adds a parallel mechanism (`Enricher.LocalizedGeoFields` + `ProcessorService.addLocalizedGeocoding` called from `processRenderJob`) which:

- runs synchronous HTTP geocoding on the render worker (currently blocks rendering of cold-cache alerts)
- clones the entire `PerLangEnrichment` outer + inner maps to avoid mutating shared state (the defensive cloning is correct, but the need only exists because we're mutating post-construction)
- duplicates the `distinctLanguages` loop that the per-type handlers already do
- pollutes base enrichment with `geocodeLatitude` / `geocodeLongitude` fields needed only because the render step lost access to lat/lon

### Suggested rework

**1. Add `addLocalizedGeoResult` as a sibling of `addGeoResult` in `internal/enrichment/enrichment.go`:**

```go
// addLocalizedGeoResult writes per-language address fields into m
// using the localized reverse-geocode result. No-op when the geocoder
// is disabled or when lang is empty (the base addGeoResult already
// covered the default-language path).
func (e *Enricher) addLocalizedGeoResult(m map[string]any, lat, lon float64, lang string) {
    if e.Geocoder == nil || lang == \"\" {
        return
    }
    addr := e.Geocoder.GetAddressForLanguage(lat, lon, lang)
    if addr == nil {
        return
    }
    m[\"addr\"] = addr.Addr
    m[\"flag\"] = addr.Flag
    m[\"streetName\"] = addr.StreetName
    m[\"streetNumber\"] = addr.StreetNumber
    m[\"city\"] = addr.City
    m[\"state\"] = addr.State
    m[\"county\"] = addr.County
    m[\"zipcode\"] = addr.Zipcode
    m[\"country\"] = addr.Country
    m[\"countryCode\"] = addr.CountryCode
    m[\"neighbourhood\"] = addr.Neighbourhood
    m[\"suburb\"] = addr.Suburb
    m[\"formattedAddress\"] = addr.FormattedAddress
}
```

**2. Call it at the end of each `*Translate` method.** Example for `PokemonTranslate` (`internal/enrichment/pokemon.go:328`):

```go
func (e *Enricher) PokemonTranslate(base map[string]any, pokemon *webhook.PokemonWebhook, lang string) map[string]any {
    if e.GameData == nil || e.Translations == nil {
        return nil
    }
    m := make(map[string]any, 30) // bumped from 20 — now also carries localized geo

    // ... existing translation work unchanged ...

    // Localized address fields. Layer 4 (perLang) of the LayeredView,
    // so these shadow the base addr/city/... when present.
    e.addLocalizedGeoResult(m, pokemon.Latitude, pokemon.Longitude, lang)

    return m
}
```

Same one-line addition in `RaidTranslate`, and the equivalent translate methods for gym, invasion, quest, lure, nest, fort-update, max-battle (whichever currently surface address fields).

**3. Delete what becomes redundant:**

- `Enricher.LocalizedGeoFields` (`internal/enrichment/enrichment.go`) — replaced by `addLocalizedGeoResult` inline
- `ProcessorService.addLocalizedGeocoding` and its call site (`cmd/processor/render.go:138`) — no longer needed
- `geocodeLatitude` / `geocodeLongitude` writes into base (`internal/enrichment/enrichment.go:122-124`) — not needed; the per-type handlers already have lat/lon from the webhook
- `floatFromAny` (`internal/enrichment/enrichment.go:159`) — only existed to read those base fields back

### Why this is better

- Single mental model: \"all per-language enrichment happens in `*Translate`\" stays true
- Geocoding moves off the render hot path onto the enrichment step (where the base lookup already runs)
- No shared-map mutation, no defensive cloning, no race-shape risk
- Single `distinctLanguages` loop per alert (the existing one in the handler)
- No base-layer field pollution

---

## Important (independent of the rework)

**1. Skip when user language equals the base fetch's language.** `distinctLanguages` returns every matched language including the operator's default locale. The base geocode already fetched with blank-lang (which most providers interpret as their default — usually English or the configured locale). After the rework, the `lang == \"\"` guard in `addLocalizedGeoResult` handles this for blank explicitly, but operators with `[general] locale = \"de\"` may still get a redundant lookup for `\"de\"`. Consider: if `e.Geocoder` tracks the language passed to its first base lookup (or you treat blank as a sentinel for the configured locale), elide that one.

**2. `geocodeLatitude` / `geocodeLongitude` pollute base when `e.Geocoder == nil`** (`internal/enrichment/enrichment.go:122-124`). Will be removed entirely by the rework above; flagging in case the rework approach isn't taken.

---

## Nits

- **`floatFromAny`** (`internal/enrichment/enrichment.go:159`) — only the `float64` writer exists; the `int*` / `float32` branches are unreachable. Drop them. Will be removed entirely after rework.
- **`google.go:67`** returns a \"google geocode: parse URL\" error for a hardcoded constant URL — unreachable error path. Drop the check or just `_ = err`.
- **API.md wording** (`API.md:9`): \"as a fallback\" is misleading. The base lookup is unconditional, not a fallback. Suggest: *\"The processor performs a base reverse-geocode lookup that populates the default-locale address fields, then a per-language lookup for each matched user language that shadows them in the per-language DTS layer.\"*
- **CLAUDE.md line 33** packs a paragraph onto one bullet. Split for consistency.
- **Unrelated whitespace churn** in `cmd/processor/render.go:80-83` (struct field tab realignment). Split out or revert; it makes the diff harder to read.

---

## Tests

- After the rework, the integration test should assert: \"a `PerLangEnrichment[de]` returned from `PokemonTranslate(base, pokemon, \\\"de\\\")` contains `addr` / `city` / `formattedAddress` resolved against the German provider response\". One test per per-type translate method that gets the change.
- Provider-level tests in this PR are good; keep them.
- Cache-key test pinning blank-language → historical key (preserves cache compat) is exactly the right pin to keep.

---

## Praise (parts that should survive the rework unchanged)

- **`Provider.Reverse(lat, lon, language)` interface change** — clean.
- **Cache key compatibility**: blank-language returns the historical key (`internal/geocoding/cache.go:73`). Test `TestCacheKeyForLanguage` pins this; upgrade doesn't invalidate existing pogreb entries.
- **Provider parameter names are correct**: Nominatim `accept-language`, Photon `lang`, Google `language` all match upstream docs.
- **Language normalization centralized** in one place (lowercase + trim), so `\"DE\"` and `\" de \"` collide on the same cache entry.

jfberry (commented):
## Follow-up: protections survive, but per-alert latency multiplies

Walked through the geocoder body to confirm what survives. Three protections do, one becomes a new concern.

### Survives (verified in `internal/geocoding/geocoder.go` after the diff)

| Protection | Status | Why |
|---|---|---|
| Concurrency limiter (default `Concurrency = 5`) | ✓ | `GetAddressForLanguage` IS the renamed body of the old `GetAddress` — same `g.sem <- struct{}{}` acquire on line 160. `GetAddress` now wraps it with `language=""`. Both code paths feed the same semaphore. |
| Circuit breaker (`FailureThreshold`, `CooldownMs`, half-open probe) | ✓ | Same code path. Cooldown + half-open state are not language-aware, which is correct — a misbehaving provider is misbehaving for all languages. |
| Two-layer cache (in-memory ttlcache + on-disk pogreb) | ✓ | Cache key is now `CacheKeyForLanguage(lat, lon, detail, lang)` (`cache.go:73`). Blank language returns the historical key so existing entries don't get orphaned. |
| Per-route metrics (`GeocodeTotal`, `GeocodeInFlight`, `GeocodeDuration`) | ✓ | All collected at the `GetAddressForLanguage` level — language is transparent to metrics. |

### New concern: per-alert latency, not throughput

The semaphore caps **throughput across the process** at 5 in-flight at any moment. It does NOT help **per-alert latency** when an alert needs multiple languages with cold cache.

The PR's `addLocalizedGeocoding` walks `distinctLanguages` **serially** on the render worker:

```go
for _, lang := range distinctLanguages(...) {
    fields := ps.enricher.LocalizedGeoFields(job.Enrichment, lang)
    ...
}
```

Each `LocalizedGeoFields` call:
1. acquires the semaphore (1 of 5)
2. issues an HTTP request to Nominatim/Photon/Google (~50-500ms typical)
3. releases the semaphore

For an alert that fans out to users speaking 3 distinct languages with all caches cold, render of that alert adds **3 × geocoder RTT** of blocked-on-IO time before the message even starts rendering. Render worker pool defaults to 4, so a burst of such alerts stalls rendering.

The canonical-pattern rework (per-language enrichment in `*Translate`) doesn't fix this on its own — it moves the same serial loop from the render worker to the webhook handler. Either approach needs the same mitigation if you want to spend the latency budget elsewhere.

### Mitigations worth considering (pick one or none)

1. **Cheapest**: accept it. Typical alerts have 1-2 distinct languages; cache is usually warm; this only bites cold-start and burst-of-new-languages cases. Just document the cost.

2. **Bounded fan-out**: replace the serial loop with a `sync.WaitGroup`-bounded concurrent loop:

```go
var wg sync.WaitGroup
results := make(map[string]map[string]any, len(langs))
var mu sync.Mutex
for _, lang := range langs {
    wg.Add(1)
    go func(lang string) {
        defer wg.Done()
        f := ps.enricher.LocalizedGeoFields(job.Enrichment, lang)
        mu.Lock()
        results[lang] = f
        mu.Unlock()
    }(lang)
}
wg.Wait()
```

The geocoder's own semaphore (`Concurrency = 5`) bounds the concurrent provider calls automatically — the fan-out can be unbounded at the caller because the geocoder back-pressures. Render time becomes `max(geocoder RTT per lang)` instead of `sum`.

3. **Pre-warm in webhook handler**: kick off language lookups when the matched users list is known, before queuing the render job. Less invasive than fanning out from the render worker but adds a dispatch step.

### Recommendation

Defer the mitigation until you see render-queue depth correlated with cold geocode bursts in production. The semaphore + cache + circuit breaker are still doing their job; the loss is per-alert latency in a narrow case, not throughput.

Move the geocoding into `*Translate` per the rework already requested — that doesn't make the latency problem any worse, and makes future fan-out cleaner (one place to add concurrency, not two).

jfberry (approved):
## Review 2: rework looks great — approving

Verified the rework end-to-end. The structural concern from the first review is resolved cleanly.

### What was reworked correctly

- ✓ **`addLocalizedGeoResult` is the new sibling** of `addGeoResult` in `enrichment.go`, exactly the canonical pattern.
- ✓ **Called from every `*Translate` method** that already exists: pokemon, raid, gym, invasion, lure, nest, quest, maxbattle, weather. New `FortUpdateTranslate` added for fort updates which previously had no translate method.
- ✓ **All the cruft is gone**: `Enricher.LocalizedGeoFields`, `ProcessorService.addLocalizedGeocoding`, `floatFromAny`, the `geocodeLatitude` / `geocodeLongitude` pollution of base — all removed. No code-smell leftovers from the original approach.
- ✓ **Skip when `lang == DefaultLocale`** — uses `e.DefaultLocale` (already on the Enricher from `cfg.General.Locale`), so a user with the operator-default language doesn't trigger a redundant provider call. Test pins this.
- ✓ **Skip when `lang == ""`** — same optimization. Test pins this implicitly.
- ✓ **Call-site updates are consistent**: every `*Translate` invocation across the per-type handlers passes `lat, lon` through. Compiler / type system enforces nothing got missed.
- ✓ **Tests cover the right invariants**: accept-language header forwarded, address fields appear in per-lang map for non-default users, no HTTP call + no per-lang shadowing for default-locale users.
- ✓ `go test ./...` and `go vet ./...` both clean.

### Praise the `defer` pattern

```go
m := make(map[string]any, 18)
defer e.addLocalizedGeoResult(m, lat, lon, lang)
```

Smart choice. The deferred call captures `m` by reference, so any early return in the Translate function still gets the address fields. Reads cleanly at the top of each method without polluting the bottom. Nice.

One stylistic nit: `FortUpdateTranslate` (the newly-added one) calls it directly without `defer`, breaking the pattern. Either move to `defer` for consistency, or note why this one is different — but it's the same shape as the others, so probably just an oversight.

### Subtle behaviour to be aware of (not blocking)

The "skip when `lang == DefaultLocale`" optimization implicitly assumes the **base** geocode (called without a language parameter) returns the operator's default-locale result. This is provider-dependent:

- **Nominatim** with no `accept-language`: returns the local language for the place (German addresses in Germany, French in France, etc.)
- **Photon**: similar
- **Google**: depends on server location / app default

So an operator with `[general] locale = "en"` running in Germany: blank-language base geocode returns German addresses; English-language users skip the per-lang lookup (because `en == DefaultLocale`); they end up seeing German addresses despite their `en` setting.

This isn't a regression — pre-PR, every user got the same blank-language base result, so cross-border setups already had this behaviour. The PR just doesn't *improve* it for users whose language matches the operator default.

The cleaner fix (future work): make `addGeoResult` (base) explicitly call `GetAddressForLanguage(lat, lon, e.DefaultLocale)` so the base lookup is genuinely the default-locale result. Then the skip is correct in all cases. One-line change but want to flag it as a follow-up rather than block this PR.

### Concurrency follow-up from review 1 still applies

The per-alert serialisation of per-language geocoder calls now happens on the webhook-handler goroutine (the per-type handler's `for _, lang := range distinctLanguages(...)` loop) instead of on the render worker. Better — webhook handlers can pile up while render moves on — but cold-cache alerts to N-language audiences still pay `N × geocoder RTT`. The semaphore (`Concurrency = 5`) caps total throughput across the process; doesn't help per-alert latency.

Bounded fan-out (snippet in my previous comment) is the future mitigation if production shows it. Defer it.

### Minor nits (none blocking)

- **Inconsistent `defer` use** noted above (FortUpdateTranslate).
- **`google.go:67`** "google geocode: parse URL" error path — was flagged in review 1, still present. Unreachable for a hardcoded valid URL. Drop the check.
- **Whitespace churn** in `cmd/processor/render.go:80-83` from the first revision is gone — thanks for cleaning that up.
- **CLAUDE.md / API.md wording** — was tweaked. Worth a re-read but no blocker.

### Verdict

**Approved.** The rework hits the canonical pattern, the deletions are complete, the tests pin the right invariants, and the latency / cross-border-locale concerns are documented for follow-up rather than blocking. Solid second pass.

PR #151, opened 2026-06-19
GeoNames intersection support

auisgold:
Adds getting the nearest street intersection from Geonames as was allowed in PoracleJS. 
Config and dts settings are the exact same as they were in PoracleJS

### DIFF
--- a/processor/cmd/processor/main.go
+++ b/processor/cmd/processor/main.go
@@ -1505,6 +1505,11 @@ func NewProcessorService(cfg *config.Config, stateMgr *state.Manager, database *
 		}
 	}
 
+	if len(cfg.Geocoding.IntersectionUsers) > 0 {
+		enricher.Intersection = geocoding.NewIntersection(cfg.Geocoding.IntersectionUsers)
+		log.Infof("GeoNames intersection lookups enabled (%d users)", len(cfg.Geocoding.IntersectionUsers))
+	}
+
 	// Stats tracker (rarity + shiny, shared rolling window)
 	statsTracker := tracker.NewStatsTracker(tracker.StatsConfig{
 		MinSampleSize:       cfg.Stats.MinSampleSize,
--- a/processor/internal/config/config.go
+++ b/processor/internal/config/config.go
@@ -647,6 +647,7 @@ type GeocodingConfig struct {
 	// Static map tile provider
 	StaticProvider    string `toml:"static_provider"`
 	StaticProviderURL string `toml:"static_provider_url"`
+	IntersectionUsers []string `toml:"intersection_users"`
 	// Optional private URL the processor uses for its own tileserver HTTP
 	// (render, pregenerate POST, upload-images prefetch). If unset,
 	// StaticProviderURL is used for everything. Useful when
--- a/processor/internal/enrichment/enrichment.go
+++ b/processor/internal/enrichment/enrichment.go
@@ -79,6 +79,7 @@ type Enricher struct {
 	RequestShinyImages bool                // Whether to request shiny icon variants
 	StaticMap          *staticmap.Resolver // Static map tile resolver (nil = disabled)
 	Geocoder           *geocoding.Geocoder // Reverse geocoder (nil = disabled)
+	Intersection       *geocoding.Intersection // Get nearest street intersection
 
 	// Fallback icon URLs when uicons are not configured or fail
 	FallbackImgURL      string
@@ -210,6 +211,13 @@ func (e *Enricher) addAddressFields(m map[string]any, addr *geocoding.Address) {
 	m["neighbourhood"] = addr.Neighbourhood
 	m["suburb"] = addr.Suburb
 	m["formattedAddress"] = addr.FormattedAddress
+}
+
+func (e *Enricher) addIntersection(m map[string]any, lat, lon float64) {
+	if e.Intersection == nil {
+		return
+	}
+	m["intersection"] = e.Intersection.GetIntersection(lat, lon)
 }
 
 // Tile mode constants. Defined here to avoid import cycles with cmd/processor.
--- a/processor/internal/enrichment/fort.go
+++ b/processor/internal/enrichment/fort.go
@@ -177,6 +177,7 @@ func (e *Enricher) FortUpdate(lat, lon float64, fortID string, fort *webhook.For
 
 	// Reverse geocoding
 	e.addGeoResult(m, lat, lon)
+	e.addIntersection(m, lat, lon)
 
 	// Static map tile — use autopositioned center if available, else original coords
 	mapLat, mapLon := lat, lon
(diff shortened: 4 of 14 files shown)


jfberry:
I assume you were using this? I left it out deliberately because I didn't think anyone used it.
I'll get claude to do a full review however I immediately notice that dts.md not updated; the dts array for the online dts editor; the config array for the online config; the example config.toml

jfberry:
## Review: GeoNames intersection support

Thanks for porting this! The approach is small and the wiring is clean. There's one must-fix before it's mergeable (an uncached, blocking external call on the enrichment hot path) plus a set of robustness/consistency items. I also diffed it against the original PoracleJS implementation (`src/lib/getIntersection.js`) since the PR says config/dts are "the exact same" — a few divergences below.

Each item is written so it can be actioned directly.

---

### 🔴 1. Must-fix: uncached, blocking HTTP on the enrichment hot path

`Enricher.addIntersection` runs inline in every enrichment path, and `Intersection.GetIntersection` does a blocking `client.Get` to `api.geonames.org` (5 s timeout) **on every webhook, with no caching**. Enrichment runs in the bounded webhook worker pool (`[tuning] webhook_workers`, default 4), so each call ties up a worker for up to 5 s, and one busy area will exhaust GeoNames credits in minutes (`findNearestIntersectionOSM` is credit-costed; the free tier is a few thousand/day).

**Fix:** cache results keyed on coordinates rounded to a few decimals, exactly like the reverse geocoder already does in this package (`internal/geocoding/geocoder.go` uses a pogreb + ttlcache layer via `CacheKeyForLanguage`/`CacheDetail`). Intersections are even more cacheable than addresses because pokestops/gyms are fixed points — a coord-rounded cache gets near-100% hit rate after warm-up. Reuse the `Cache`/`CacheKey` machinery (hold a `*Cache` on the `Intersection` struct, or fold the lookup into the geocoder).

> Note: PoracleJS is *also* uncached, so the logic is a faithful port — but the runtime behavior is not. In Node, `await axios.get(...)` yields the single-threaded event loop (non-blocking); in Go, `client.Get` blocks a pooled goroutine. So the same logic is materially worse here, which is why caching is required rather than optional.

---

### 🟠 2. Endpoint was changed vs PoracleJS — not the "exact same" data source

- PoracleJS: `findNearestIntersection**JSON**` (US-only, Census TIGER)
- This PR: `findNearestIntersection**OSM**JSON` (OpenStreetMap, broader coverage)

Different GeoNames endpoints, different coverage and credit cost. This looks like a deliberate improvement (worldwide vs US-only), which is great — but it contradicts the "exact same as PoracleJS" claim and changes the response assumptions. **Please confirm it's intentional and call it out in the PR/docs** (it's a selling point for non-US operators).

---

### 🟠 3. Observability regression vs PoracleJS

PoracleJS logs failures at **`log.warn`** and surfaces the GeoNames `result.data.reason` (which carries the over-quota message). GeoNames returns **HTTP 200** with a `{"status":{"value":18/19/20,"message":...}}` body when hourly/daily credits are exhausted — so the current code (which only checks `StatusCode != 200`, then reads `street1`/`street2`) silently degrades to `"No intersection"` forever once you're rate-limited, and only logs at `Debugf`.

**Fix:** decode the `status` object and log at `Warnf` with the GeoNames message (matching PoracleJS), so operators can distinguish "no intersection here" from "we're throttled."

---

### 🟡 4. Failure sentinel is inconsistent and breaks PoracleJS template compat

`GetIntersection` returns `"No intersection"` (lowercase i) on failure but `""` on the no-users path. PoracleJS returns `'No Intersection'` (capital I) **consistently on every failure and empty path**. So this is inconsistent with itself *and* a casing break — any ported template doing `{{#eq intersection "No Intersection"}}` silently stops matching.

**Fix — pick one deliberately:**
- Faithful port: return `"No Intersection"` everywhere.
- Intentional improvement: return `""` everywhere, so templates can use `{{#if intersection}}…{{/if}}` to hide the line.

Either is fine; just be consistent and note the choice.

---

### 🟡 5. Dead branch

The `len(i.Usernames) == 0` guard in `GetIntersection` is unreachable: `addIntersection` only calls it when `e.Intersection != nil`, and `Intersection` is only constructed when `len(IntersectionUsers) > 0` (`main.go`). Drop it, or keep it as the single empty-return path per item #4.

---

### 🟡 6. Per-call RNG and `http.Client` allocation

```go
randSource := rand.NewSource(time.Now().UnixNano())
r := rand.New(randSource)
username := i.Usernames[r.Intn(len(i.Usernames))]
```

A fresh time-seeded RNG per call is wasteful and, under concurrency, calls in the same nanosecond window pick correlated usernames. PoracleJS just uses `Math.random()` (global). **Fix:** use the global `rand.Intn(len(i.Usernames))` (auto-seeded and mutex-safe on Go 1.20+), or seed once in `NewIntersection`. Likewise, build one `*http.Client` in `NewIntersection` and reuse it (connection keep-alive) instead of allocating per request.

---

### 🟡 7. Use HTTPS + a context

`http://api.geonames.org` sends the username in cleartext (the original also used `http`, but `https` is supported — worth upgrading). And `client.Get` ignores cancellation; prefer `http.NewRequestWithContext` so a shutting-down/over-deadline pipeline can abort the call.

---

### 🟡 8. Missing trailing newline + no tests

`intersection.go` ends without a trailing newline (`\ No newline at end of file`) — gofumpt / `golangci-lint` will flag it. There are also no tests; the parsing logic (`street1 & street2` → `"A & B"`, partial/missing → sentinel, over-quota `status` body → sentinel) is easy to cover with an `httptest.Server`.

---

### 🟡 9. Plumbing/docs — make the field discoverable in **every** DTS type

The field should be available in all DTS types, so register it as a **common** field (not per-type):

- **DTS editor field list:** add an `intersection` entry to `commonFields` (the `Category: "location"` block, near `addr` at `processor/internal/api/dts_fields.go:28`). Putting it in `commonFields` makes `/api/dts/fields/{type}` advertise it for every type automatically.
- **DTS.md:** add an `intersection` row to the location-fields table (next to `addr`/`streetName`, ~line 201).
- **config.example.toml:** add `intersection_users` under `[geocoding]` (~line 403) with a comment and a GeoNames-credits caveat.
- **Config editor schema/values:** add `intersection_users` so the online config UI exposes it.

---

### Meta
Like external nominatim, this exfiltrates every spawn coordinate to a third party (GeoNames). Fine, but worth one line in the docs.

**Summary:** item #1 (uncached blocking call) is the blocker; #2/#3 are the meaningful divergences from the original; the rest is consistency and polish. Happy to help with the cache integration if useful.

jfberry:
I'm happy to take this forward if you'd like to use it - I am not sure if the change to data source is significant or not?

auisgold:
Would be awesome if you could help with the changes needed. I changed the data source as I started covering non-US locations, but as far as I know there is no difference in cost.

jfberry:
See #154 which claude has had a go at - lmk if it works and I will merge it

jfberry:
Superceded by #154 which is now merged to develop

PR #154, opened 2026-06-23
geocoding: cached GeoNames intersection lookups ({{intersection}}) — takeover of #151

jfberry:
Takes over #151 (thanks @auisgold for the original port) and addresses the review feedback there. Keeps the OSM endpoint change (worldwide coverage) as agreed.

## What this adds
A `{{intersection}}` DTS field populated from GeoNames `findNearestIntersectionOSM` — the nearest street intersection (`"Street1 & Street2"`) for an alert's coordinates. Registered as a **common** field, so it's available on every DTS type.

## The headline fix: caching, shared with the geocoder
The original port did an **uncached, blocking** GeoNames request on the enrichment hot path for every webhook. This now shares the reverse-geocoder's pogreb cache.

pogreb is a single key-value store with no notion of tables, so the answer to "can it have more than one table?" is: **namespace the keys in the same DB**. Intersection entries live in the same on-disk pogreb file as addresses but under an `"isect:"` key prefix (`IntersectionCacheKey`), with their own in-memory layer — effectively a second logical table in one file. Pokestops/gyms are fixed coordinates, so the coord-rounded cache reaches near-100% hit rate after warm-up.

- **Negative caching:** a stable "no intersection here" is cached (so rural stops don't re-query forever); transient failures are **not** cached, so they retry on the next sighting.
- **Sharing:** when a reverse-geocode provider is configured, the cache is shared automatically. If geocoding is `none` there's no geocache to share, so intersection runs uncached (still concurrency-bounded) — `intersection_users` without a geocoder is the only uncached path.

## Other review corrections
- Reuse one `http.Client`; bound outbound calls with a semaphore so a cold cache can't exhaust the webhook worker pool (Go blocks a goroutine where PoracleJS's `await` yielded the event loop).
- Global `rand` instead of a per-call time-seeded source.
- HTTPS + `context`-with-timeout instead of a bare `http.Get`.
- Decode GeoNames' HTTP-200 error body (e.g. credit exhaustion) and log at **Warn** with the reason, instead of silently degrading to a sentinel.
- Return `""` consistently on every failure/empty path so templates can do `{{#if intersection}}`. **Note:** this diverges from PoracleJS, which returned the literal `"No Intersection"`; chose `""` to match this codebase's empty-on-failure geocoding convention.

## Plumbing the original PR was missing
- `intersection` in `commonFields` (`internal/api/dts_fields.go`) → advertised for every type by `/api/dts/fields/{type}`
- `DTS.md` location-fields table
- `[geocoding] intersection_users` in `config.example.toml`
- geocoding section of the online config-editor schema (`internal/api/config_schema.go`)

## Tests
- `cache_test.go`: intersection round-trip, negative cache (empty = hit), address/intersection namespace isolation
- `intersection_test.go` (httptest): street parsing, empty-on-no-streets, GeoNames error body, HTTP 500, success caching (1 HTTP call), negative caching, failures-not-cached, no-usernames no-op, timeout

Full gate green: `go build ./... && go vet ./... && go test -count=1 ./... && golangci-lint run ./...`

Closes #151.

### DIFF
--- a/DTS.md
+++ b/DTS.md
@@ -210,6 +210,7 @@ These fields are available in every template:
 | `neighbourhood` | string | Neighbourhood name |
 | `suburb` | string | Suburb name |
 | `flag` | string | Country flag emoji |
+| `intersection` | string | Nearest street intersection (`Street1 & Street2`) from GeoNames; empty when disabled (`[geocoding] intersection_users`) or none nearby |
 | `staticMap` | string | Static map tile image URL |
 | `staticmap` | string | *Deprecated* — alias for `staticMap` |
 | `imgUrl` | string | Primary icon URL |
--- a/config/config.example.toml
+++ b/config/config.example.toml
@@ -413,6 +413,12 @@ provider_url = ""                               # nominatim or photon URL, e.g. 
 # own template like "{{{streetName}}} {{streetNumber}}, {{{city}}}".
 forward_only = false                     # when true, disable reverse geocoding lookup
 cache_detail = 3                         # decimal places of lon/lat for caching (3 or 4 for 100x more detail)
+# intersection_users - list of GeoNames usernames (https://www.geonames.org)
+#   used to populate the {{intersection}} DTS field with the nearest street
+#   intersection. One is picked at random per uncached lookup to spread credit
+#   usage. Empty disables the feature. Results share the geocoding cache above,
+#   but each uncached lookup spends a GeoNames credit, so caching/credits apply.
+intersection_users = []
 # Styles for different times of day using tileservercache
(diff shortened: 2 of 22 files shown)


auisgold:
In order to use HTTPS, endpoint needs to be switched to https://secure.geonames.org/findNearestIntersectionOSMJSON instead. Other than that, everything seems to work fine.

jfberry:
Good catch — fixed in 1f8c69cf. Switched the default endpoint to `https://secure.geonames.org/findNearestIntersectionOSMJSON`; `api.geonames.org` doesn't answer over TLS, so the `https://` scheme needed the `secure.` host. Thanks for testing it end-to-end.

PR #161, opened 2026-07-14
feat(showcase): ingest, alert on, and describe pokestop-webhook showcases (Phase 1+2)

jfberry:
Implements **Phase 1** of the showcase design (`docs/superpowers/specs/2026-07-14-showcase-support-design.md`).

## Problem
Showcases (pokéstop contests) produced no correct alert. Verified against the Golbat source: the **`pokestop`** webhook carries the showcase *content* (`showcase_focus`, `showcase_expiry`, `showcase_rankings`) but **no `display_type`**; the **`invasion`** webhook fires with `display_type=9` but is a bare, content-less envelope. PoracleNG's showcase support keys on the invasion/`display_type>=7` path — so the content-rich pokéstop webhook was dropped into `ProcessInvasion` as a degenerate `grunt_type=0` invasion that matched nobody, and the content-less invasion webhook would render an empty card. Neither alone renders a showcase.

## Approach
Drive showcases from the `pokestop` webhook and **reuse the incident downstream** by synthesising `display_type=9`. **No new tracking type** — a showcase is an `/incident` rule (`grunt_type="showcase"`) in the existing `invasion` table, rendered by the `incident` template. (Structure decision + reversibility documented in the design doc.)

- **`ShowcaseWebhook`** parsing (`webhook/types.go`).
- **`routePokestop`** dispatches showcase, lure, and invasion **independently** (the pokéstop webhook is a multiplexed snapshot); a showcase-only stop no longer falls through the legacy invasion fallback.
- **`ProcessShowcase`** (`cmd/processor/showcase.go`): gate on `showcase_expiry > now` (Golbat never clears these fields, so stale showcases linger on lure/power-up fires — same class as the lure gate in #160); dedup on `(pokestop, expiry, rank-1 fingerprint)`; synthesise `display_type=9`; reuse the invasion matcher + incident enrichment/template.
- **Edit-mode** wired: `EditKey="showcase:<pokestop>:<expiry>"` + `OverrideCleanTTH=showcase_expiry`, reusing the raid-RSVP path. A contest first fires with an empty leaderboard then re-fires on rank-1 movement; with the edit bit, all fires edit one message in place.
- **Suppress** the content-less `display_type=9` invasion envelope in `ProcessInvasion` so it can't produce a duplicate empty card (kept as a safety net — operator couldn't confirm whether their scanner emits it).
- **`disable_showcase`** config toggle; `poracle-test showcase,{type,buddy}` wiring + testdata.

## Tests
- Routing dispatch via a recording mock (`receiver_test.go`): showcase-only → `ProcessShowcase` and **not** the invasion fallback; lure+showcase → both; real invasion unaffected.
- `showcaseActive` gate (future vs past/absent expiry) + `showcaseFingerprint` (rank-1 sensitive, `last_update`/rank-2 insensitive).
- `CheckShowcase` dedup (repeat state dedups; rank-1 change passes through; new contest passes through).
- Invasion suppression wiring.
- **Payoff verification** (`enrichment/invasion_test.go`): fed exactly what `ProcessShowcase` passes (gruntTypeID=0 + `showcase_rankings`), the enrichment populates `showcasePresent` + the translated top-3 leaderboard + winner.

Full gate green: `go build`, `go vet`, `go test -count=1 ./...`, `golangci-lint run ./...` (0 issues).

## Deliberately out of scope (per operator decisions, see doc)
- **Phase 2**: `showcase_focus` "what's featured" enrichment (translated type/pokémon/class names) + DTS fields + default template polish.
- Focus-based tracking filters (match-all only — no dedicated showcase structure yet, pending user feedback).


jfberry:
## Phase 2 added (focus enrichment)

Pushed the "what's featured" enrichment on top of Phase 1, so showcase alerts now describe the contest, not just the leaderboard.

- Parses `showcase_focus` (all 10 focus classes) → `showcaseFocusType` / `showcaseFocusCategory` (translated) / `showcaseFocusName` (translated value) / `showcaseFocusEmoji`.
- Focus classes enumerated in `util.json` `showcaseFocus`; category labels in translation (`showcase_focus_{type}`) — the data-driven pattern.
- **Value enums mapped explicitly from the game proto** (verified against `Golbat decoder/pokestop_showcase.go`), because two don't map by index:
  - `alignment` is **reversed** vs gamelocale (`ContestPokemonAlignmentFocusProto` 1=PURIFIED/2=SHADOW, but `alignment_1`=Shadow/`alignment_2`=Purified).
  - `generation` proto (`GEN8A=9` Hisui, `GEN9=10` Paldea) doesn't line up with the 1–9 gen numbering.
  - `class` = normal/legendary/mythic/ultra-beast (0–3); type/pokemon/family reuse existing keys.
- DTS field metadata + `DTS.md` docs; tests cover the alignment (purified/shadow) and generation (proto 10 → Paldea) mappings.

Design doc updated (Phase 1 + 2 done; only minor template polish left for Phase 3). Full gate green.

PR #169, opened 2026-07-22
fix(rowtext): show explicitly-tracked Normal form in !tracked

jfberry:
## Problem
A user tracking `rattata form:normal` (and other forms) sees duplicate-looking rows in `!tracked` with no way to tell them apart:

```
[id:305756] Rattata Alola | ...
[id:305757] Rattata       | ...   ← this is the Normal form
[id:305758] Rattata       | ...
```

## Root cause
`translateMonsterName` (`internal/rowtext/rowtext.go`) suppressed **any** form whose translated name was `"Normal"`/`"Unset"`/`""`, regardless of the form value:

```go
} else if enrichment.IsNormalForm(formName) {   // fires for ANY form value
    formName = ""
}
```

Rattata's Normal form is stored as a non-zero value (form **45**; masterfile `defaultFormId: 45`, `forms: [45, 46]`), and `form_45` → `"Normal"` from gamelocale. So the explicit Normal form rendered identically to the `form 0` ("any form") tracking. Alola (46) survived only because "Alola" isn't a Normal-ish name.

## Fix
Only `form 0` ("any form") drops a default Normal/Unset label. An explicitly-tracked form (`form != 0`) keeps its name — including "Normal" — so distinct form trackings stay distinguishable:

```
[id:305756] Rattata Alola  | ...
[id:305757] Rattata Normal | ...
[id:305758] Rattata        | ...   (form 0 = any)
```

Scoped to `rowtext` (tracking descriptions); alert/enrichment display uses a separate path, so no "Normal" noise is added to actual alerts.

## Tests
- Added a failing-first table-driven `translateMonsterName` test (normal / named / any-form).
- Full gate green: `go build ./... && go vet ./... && go test ./... && golangci-lint run ./...`.

### DIFF
--- a/processor/internal/rowtext/rowtext.go
+++ b/processor/internal/rowtext/rowtext.go
@@ -59,8 +59,10 @@ func translateMonsterName(tr *i18n.Translator, gd *gamedata.GameData, pokemonID,
 		// Translation key returned as-is means no translation found
 		formName = ""
 	} else if form == 0 && enrichment.IsNormalForm(formName) {
-		formName = ""
-	} else if enrichment.IsNormalForm(formName) {
+		// form 0 = "any form": suppress a default "Normal"/"Unset" label so the
+		// row reads as the whole species. An explicitly-tracked form
+		// (form != 0) keeps its name — including "Normal" — so distinct form
+		// trackings stay distinguishable in !tracked.
 		formName = ""
 	}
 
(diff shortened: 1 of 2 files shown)


kbtbc:
Thanks for the fix.  Similarly in the new 'costume' tracking, there is no real differentiator in the tracked output between a costume and a form tracking: 

```
Pikachu  | Flower | distance: 5000m ...
Pikachu Doctor | distance: 5000m  ... 
```

The top one is a `costume:flower` tracking, the other is a `form:doctor` tracking.

PR #182, opened 2026-08-09
fix(delivery): route clean-deletion through the FairQueue (fix 429 storm)

jfberry:
## Problem
With clean tracking across 20+ areas, expired raids/eggs deleting together hit Discord's DELETE rate limit and **failed after 6 retries** — the message then stayed in the channel:
```
WARN [clean-delete] discord: 429 for DELETE 1377…, retry_after=0.6s (attempt 3/6)
WARN [clean-delete] discord: 429 for DELETE 1377…, retry_after=0.5s (attempt 5/6)
…
WARN delivery: clean delete failed for 1377…: discord delete returned status 429
```

## Root cause
**Clean-deletion had no concurrency control.** Each TTL eviction spawned its own `go sender.Delete`, so a burst of expiring alerts fired *dozens of concurrent DELETEs to the same channel*. They 429'd each other and each independently burned through its retries, then gave up. (The logs show many "attempt N/6" for one channel in the same second.) Adding retry alone — my first commit — doesn't help when they're all racing.

## Fix (per the report: same queue, different job type)
Route clean-deletion **through the FairQueue as a delete job**, so it takes the same **per-destination lock** that already serialises sends (max 1 in-flight per target) + `WaitForRateLimit` + the platform semaphore. Deletes to a channel now run one at a time, respecting the shared bucket, instead of racing.

- `Job.DeleteSentID` marks a clean-delete job; `FairQueue.processJob` handles it under the per-destination lock, calls `Sender.Delete`, and skips alert-limit accounting / tracking / snapshot.
- `MessageTracker.cleanDelete` routes through a `cleanDeleteHook` the Dispatcher wires (`enqueueCleanDelete`) — enqueues the delete `Job`. Set before `Load()` so recovery-time cleans route through it too. Non-blocking + panic-guarded (a delete dropped mid-shutdown / on a saturated queue is re-attempted on the next startup load). Falls back to the direct-Delete goroutine when no queue is wired (tests).
- `DiscordSender.Delete`/`Edit` (and Telegram `deleteMessage`/`Edit`) keep the reactive **429 Retry-After backoff** from the first commit (`doWithRetry` / `doPostWithRetry`). `Delete` no longer self-`Wait`s — the queue does it now.

## Two commits
1. `rate-limit + retry clean-delete and edit (429 backoff)` — reactive retry (necessary but insufficient on its own).
2. `route clean-deletion through the FairQueue (serialise per destination)` — the actual fix: serialisation.

## Testing
- `TestFairQueue_CleanDeleteSerializedPerTarget` — 20 deletes to one channel with `ConcurrentDiscord=8` serialise to **max 1 concurrent** (and a delete job never triggers a Send).
- `TestTracker_CleanDeleteRoutesThroughHook` — eviction routes through the hook, not the direct-Delete fallback.
- `TestDiscord/TelegramDelete_RetriesOn429` — 429 → retry → success.
- All existing send/edit/queue/tracker tests green; `-race` clean; full gate green (build/vet/test/lint 0 issues).
- CLAUDE.md delivery rate-limiting section updated.


jfberry:
Superseded by #183, which carries the full delivery-queue redesign (per-destination lanes + semaphore-held-only-during-the-wire-call) and absorbs this PR's clean-delete commits (811f254a, 4fd21365). The 429 clean-delete storm and the shared-queue global-degradation problem are both fixed there. Closing this in favor of #183.

PR #188, opened 2026-08-16
perf: cut processor RSS 6.4x, 6436 MB -> 1000 MB (bucketed rarity ring + dedup fingerprints + tzf DefaultFinder)

TurtIeSocks:
## Summary

A production install running ~2400 pokemon webhooks/sec sat at **6436.5 MB RSS on `develop`, climbing until it OOM'd**. On this branch the same box holds **~1000 MB, steady**, through a community day at the same rate. That is a **6.4x cut in resident memory**, and 9.1x in live heap.

A heap profile showed why: 3777.80 MB of live heap, **92% of it in two structures on the pokemon path**. Both scale with webhook rate rather than with users or tracking rules, so they grow on their own with no config change.

Four fixes, one commit each. Three of them are the same mistake in different clothes: keeping raw per-event material when only an aggregate was ever read back.

Same host, `inuse_space`, before and after:

```
develop (3777.80MB total)              this branch (416.50MB total)
1838.64MB 48.67%  RecordSighting         3.04MB  0.73%  RecordSighting
1643.72MB 43.51%  CheckPokemon         299.70MB 71.96%  expiringSet.Add
 140.91MB  3.73%  geo.init.0            28.66MB  6.88%  geo.init.0
```

## 1. Rarity: per-sighting slice → bucketed counter ring

`StatsTracker.RecordSighting` appended a 24-byte `sighting` record per pokemon webhook — **before** the duplicate check (`pokemon.go:55` vs `:67`), so duplicates counted too — and kept `window_hours` of them. At 2400/sec over the default 8h window that is ~70M records.

Those records were only ever read to produce **per-species counts** in `recalculate()`. So store the counts: a ring of per-minute buckets, each `map[int32]{all, iv, shiny}`. Size now depends on distinct species and window length, never on throughput.

The second half of the bug is what made it fatal. Pruning used `pruned := st.sightings[:0]`, which reuses the backing array, and **Go slices never shrink capacity**. The high-water mark was therefore permanent: every spawn surge ratcheted the floor up and it only came back on restart. That also explains the overnight growth pattern in the profiles, where `RecordSighting` grew 25% while the dedup cache (which really evicts) grew 15%.

Window granularity moves from one second to one minute. The clock is now injectable so window expiry is testable without sleeping.

## 2. Dedup: `ttlcache[string, bool]` → sharded fingerprint map

`DuplicateCache` spent **265 B per entry** (measured) to remember one bit: a heap-allocated `Item`, the retained key string, a `container/list` node, and an `expirationQueue` entry. The old profile shows the shape plainly — 267.51 MB of nothing but list nodes, and 146 MB of nothing but the `fmt.Sprintf` that built keys the cache then held onto.

The new `expiringSet` (`internal/tracker/expiring_set.go`) hashes the key with `maphash` and stores only its expiry, in a map sharded 16 ways. **37.8 B/entry**: 126.4 MB → 18.0 MB over 500k entries in a standalone benchmark.

**Key derivation and TTLs are deliberately untouched.** Every `Check*` method builds the same key string (CP still in the pokemon key) and passes the same TTL, so dedup semantics are identical. Reads expire lazily exactly as `ttlcache` did. A background sweeper reclaims expired entries **one shard at a time**, because a full unsharded scan of a multi-million-entry map measured 89 ms and would have traded memory for a latency spike on the webhook path.

`raidCache` stays on `ttlcache` — it stores real values (prior RSVP state) rather than a bit, and its cardinality is bounded by active raids.

Two behaviour notes. Keys are fingerprinted rather than retained, so a 64-bit collision reads as a false duplicate and drops one alert; at ~6M live entries that is ~3.3e-13 per insert, roughly one every few decades at 2400/sec. And expiry is stored in unix **nanoseconds**, not seconds, because seconds truncate a sub-second TTL into the past on write. The expiry test caught that one.

## 3. tzf: `NewFullFinder()` → `NewDefaultFinder()`

`geo.init.0` decoded tzf-dist's **18 MB full-precision** polygon set at startup. `NewDefaultFinder` decodes the **6.5 MB topology-simplified** set and shares the same preindex tile fast path. Measured against `tzf v1.2.5` / `tzf-dist v0.0.2026-c-fix1` in isolation: **163.0 MB → 33.6 MB**. In-process: **140.91 MB → 28.66 MB**.

The cost is boundary precision. Upstream bounds simplified boundaries to **~111 m** of the true border, so a spawn that close to a timezone line resolves to a neighbouring zone. The coordinate guard goes from 9 cases to 21, adding zones that sit close to a neighbour or carry an unusual offset (Phoenix, Indianapolis, Kathmandu, Adelaide, Lisbon/Madrid, Anchorage, Honolulu, Mumbai). All 21 resolve identically before and after the swap.

## 4. Weather: cells were never evicted

`grep -c 'delete(' internal/tracker/weather.go` returned **0**. Two things grew. `controllerCellData.hourWeather` gained an entry per cell per hour and kept it forever, unreadable past the previous hour (`UpdateFromWebhook` only compares against `currentHour-3600`, and `ExportCellWeather` filters to the same). Separately, a cell that stopped being scanned kept its struct indefinitely, so a shifted scan area stranded state.

A sweep on a 10-minute ticker now drops hourly entries behind the retention floor and reclaims cells untouched for 24h. The idle window is deliberately longer than the 8h AccuWeather refresh so an actively-forecast cell is never reclaimed, and `lastSeen` is stamped **before** the early return in `CheckWeatherOnMonster` so any scanned cell stays alive. Forecast hours written ahead by `SetHourWeather` are preserved.

This one is too slow to show up in a snapshot profile, so it rests on its tests rather than on the numbers below.

## Measurements

A/B'd against a shadow instance on a second VPS taking the same Golbat firehose, with an empty database and no tokens. Traffic parity was confirmed from `/metrics` over a 60s window: **2425.9 pokemon/sec on both**, ratio 1.00 across pokemon, invasion and raid.

Production, same host, before and after:

| | `develop` | this branch |
|---|---:|---:|
| `process_resident_memory_bytes` | 6436.5 MB | **~1000 MB** |
| `go_memstats_heap_inuse_bytes` | 5520.8 MB | 590-680 MB |
| `go_memstats_next_gc_bytes` | 7497.9 MB | ~811 MB |
| `inuse_space` total | 3777.80 MB | 416.50 MB |

`next_gc` at 7497.9 MB is the `GOGC=100` doubling made explicit: no collection until the heap reached 7.5 GB, so every byte of live heap removed takes about two bytes of RSS ceiling with it.

The after figures are at steady state, not a lucky sample. RSS held between 999.0 MB and 1012.3 MB across readings from 2.7h to 3.5h uptime at a flat 2419.7 pokemon/sec, and `expiringSet.Add` read 299.70 MB in two profiles 50 minutes apart. `RecordSighting` is still filling toward the 8h window and should settle in the single-digit MB.

One property worth stating rather than discovering later: Go maps do not shrink on `delete`, so `expiringSet` is bounded by peak occupancy rather than current. At constant traffic that is the same number. A sustained spike would raise the floor until restart, in the same way the old rarity slice did, though at roughly a seventh of the bytes per entry. If that turns out to matter in practice, the sweeper is the natural place to rebuild an undersized shard.

## Also included

`fix(test): pin summary scheduler tests to UTC` — an unrelated pre-existing failure the gate surfaced. The test humans sit at lat/lon 0/0, so `isScheduleActive` falls back to the host's `time.Local` when `schedulerConfig.DefaultTimezone` is empty, while every `now` in the file is a fixed UTC instant. `TestSummaryScheduler_Tick_FiresWhenScheduleMatches` and `..._RespectsScheduleWindow` therefore passed on a UTC runner and failed under any real local offset. Verified green under UTC, America/New_York, Asia/Tokyo and Australia/Sydney.

## Test

Each new test was written before its implementation and verified to fail against the old code, or against a deliberately sabotaged version where it guards pre-existing behaviour:

- `TestStatsTrackerMemoryIsIndependentOfSightingVolume` — 51.6 MB for 2M sightings on `develop`, against a 30 MB budget
- `TestStatsTrackerDropsSightingsOutsideWindow`, `TestStatsTrackerRingRecyclesStaleBuckets` — window expiry and ring wrap-around, neither of which the slice implementation ever had
- `TestDuplicateCacheMemoryPerEntry` — 265.0 B/entry on `develop`, against a 60 B budget
- `TestDuplicateCacheEntryExpires`, `TestExpiringSetSweepReclaimsExpiredEntries`, `TestExpiringSetSpreadsAcrossShards`
- `TestWeatherTrackerEvictsHoursOutsideRetention`, `TestWeatherTrackerEvictsIdleCells`
- `TestGetTimezone_KnownCoordinates` widened from 9 to 21 cases

Gate green: `go build ./... && go vet ./... && go test -count=1 ./... && golangci-lint run ./...` (0 issues).

## Open question for the maintainer

Dedup is now the largest single consumer by a wide margin — 299.70 MB, 72% of live heap — and it catches **2.3 duplicates/sec out of 2425.9 received, or 0.09%**. The flat 3600s TTL for unverified webhooks (`duplicate.go:57`) is the biggest remaining lever on it, and the pokemon key includes CP, so one spawn occupies at least two entries (the wild sighting at `cp=0`, then the encounter). Both were left alone here on purpose, since they are dedup-correctness calls rather than memory ones. Happy to follow up if either should change.

### DIFF
--- a/processor/internal/geo/timezone.go
+++ b/processor/internal/geo/timezone.go
@@ -10,7 +10,15 @@ var finder tzf.F
 
 func init() {
 	var err error
-	finder, err = tzf.NewFullFinder()
+	// NewDefaultFinder, not NewFullFinder: both wrap the same preindex tile
+	// fast path, but Full decodes tzf-dist's 18 MB full-precision polygon set
+	// where Default decodes the 6.5 MB topology-simplified one. That cost
+	// 140.96 MB of live heap at init in production (4.44% of the total) for
+	// accuracy we cannot use — upstream bounds the simplified boundaries to
+	// ~111 m of the true border, and a pokemon spawn that close to a timezone
+	// line resolves to a neighbouring zone with the same wall clock in every
+	// case a Poracle alert cares about.
+	finder, err = tzf.NewDefaultFinder()
 	if err != nil {
 		panic("failed to initialize timezone finder: " + err.Error())
 	}
--- a/processor/internal/geo/timezone.go
+++ b/processor/internal/geo/timezone.go
@@ -13,11 +13,20 @@ func init() {
 	// NewDefaultFinder, not NewFullFinder: both wrap the same preindex tile
 	// fast path, but Full decodes tzf-dist's 18 MB full-precision polygon set
 	// where Default decodes the 6.5 MB topology-simplified one. That cost
-	// 140.96 MB of live heap at init in production (4.44% of the total) for
-	// accuracy we cannot use — upstream bounds the simplified boundaries to
-	// ~111 m of the true border, and a pokemon spawn that close to a timezone
-	// line resolves to a neighbouring zone with the same wall clock in every
-	// case a Poracle alert cares about.
+	// 140.96 MB of live heap at init in production, 4.44% of the total.
+	//
+	// The trade is real, not free. Upstream bounds the simplified boundaries
+	// to ~111 m of the true border, but a neighbouring zone is not always on
+	// the same wall clock: India/Nepal differ by 15 min, the Arizona line by
+	// an hour for half the year, and China's western border by up to 2.5 h.
+	// A scan of the India/Nepal border alone finds ~65 points inside the band
+	// where the two finders disagree on the offset.
+	//
+	// A spawn or saved user location inside such a band therefore renders
+	// despawn and hatch times, and profile/summary schedules, off by that
+	// offset. The band is ~111 m wide against S2 level-10 weather cells and
+	// typical scan areas, so this is accepted rather than unnoticed: see
+	// TestGetTimezone_SimplifiedBoundaryOffsets for a pinned example.
 	finder, err = tzf.NewDefaultFinder()
 	if err != nil {
 		panic("failed to initialize timezone finder: " + err.Error())
--- a/processor/internal/tracker/expiring_set.go
+++ b/processor/internal/tracker/expiring_set.go
@@ -1,0 +1,135 @@
+package tracker
+
+import (
+	"hash/maphash"
+	"sync"
+	"time"
+)
+
+// expiringSetShards is the number of independently-locked shards. Sweeping
+// takes one shard's lock at a time, so a full sweep of a multi-million-entry
+// set never blocks the webhook path for more than a fraction of the total
+// scan.
+const expiringSetShards = 16
+
+// expiringSweepInterval is how often the background sweeper reclaims expired
+// fingerprints. Reads expire lazily regardless, so this only bounds memory,
+// never correctness.
+const expiringSweepInterval = time.Minute
+
+// expiringSet is a set of key fingerprints with per-entry expiry.
+//
+// It exists because storing dedup keys in a ttlcache cost ~265 B per entry —
+// a boxed item, the retained key string, a container/list node and an
+// expiry-heap entry — to remember a single bit. At scanner throughput the
+// live set runs to millions of keys, which measured 1.42 GB (44.8% of live
+// heap) in a production profile. Hashing the key to a uint64 and storing only
+// its expiry brings that to ~25 B per entry.
+//
+// Keys are fingerprinted, not retained, so a 64-bit collision would read as a
+// false duplicate and drop one alert. Against a live set of ~6M entries that
+// is ~3.3e-13 per insert, or roughly one dropped alert every few decades at
+// production insert rates.
+type expiringSet struct {
+	seed   maphash.Seed
+	shards [expiringSetShards]expiringShard
+	stop   chan struct{}
+	once   sync.Once
+}
+
+type expiringShard struct {
+	mu sync.Mutex
+	// entries maps a key fingerprint to its expiry in unix nanoseconds.
+	// Nanoseconds rather than seconds so a sub-second TTL is not truncated
+	// into the past on write; the value is 8 bytes either way.
+	entries map[uint64]int64
+}
+
+func newExpiringSet() *expiringSet {
+	s := &expiringSet{
+		seed: maphash.MakeSeed(),
+		stop: make(chan struct{}),
+	}
+	for i := range s.shards {
+		s.shards[i].entries = make(map[uint64]int64)
+	}
+	go s.sweepLoop()
+	return s
+}
+
+func (s *expiringSet) fingerprint(key string) uint64 {
+	return maphash.String(s.seed, key)
+}
+
+func (s *expiringSet) shardFor(fp uint64) *expiringShard {
+	return &s.shards[fp%expiringSetShards]
+}
+
+// Has reports whether key is present and unexpired. Entries past their expiry
+// read as absent even when the sweeper has not reclaimed them yet.
+func (s *expiringSet) Has(key string) bool {
+	fp := s.fingerprint(key)
+	sh := s.shardFor(fp)
+
+	sh.mu.Lock()
+	defer sh.mu.Unlock()
+
+	expiry, ok := sh.entries[fp]
+	return ok && expiry > time.Now().UnixNano()
+}
+
+// Add records key for the given ttl, replacing any existing expiry.
+func (s *expiringSet) Add(key string, ttl time.Duration) {
+	fp := s.fingerprint(key)
+	sh := s.shardFor(fp)
+
+	expiry := time.Now().Add(ttl).UnixNano()
+
+	sh.mu.Lock()
+	sh.entries[fp] = expiry
+	sh.mu.Unlock()
+}
+
+// Len returns the number of entries still held, expired or not.
+func (s *expiringSet) Len() int {
+	var n int
+	for i := range s.shards {
+		s.shards[i].mu.Lock()
+		n += len(s.shards[i].entries)
+		s.shards[i].mu.Unlock()
+	}
+	return n
+}
+
+// Close stops the background sweeper. Safe to call more than once.
+func (s *expiringSet) Close() {
+	s.once.Do(func() { close(s.stop) })
+}
+
+func (s *expiringSet) sweepLoop() {
+	ticker := time.NewTicker(expiringSweepInterval)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-s.stop:
+			return
+		case <-ticker.C:
+			s.sweep(time.Now().UnixNano())
+		}
+	}
+}
+
+// sweep drops expired entries, taking one shard lock at a time. now is unix
+// nanoseconds.
+func (s *expiringSet) sweep(now int64) {
+	for i := range s.shards {
+		sh := &s.shards[i]
+		sh.mu.Lock()
+		for fp, expiry := range sh.entries {
+			if expiry <= now {
+				delete(sh.entries, fp)
+			}
+		}
+		sh.mu.Unlock()
+	}
+}
--- a/processor/internal/tracker/expiring_set.go
+++ b/processor/internal/tracker/expiring_set.go
@@ -2,6 +2,7 @@ package tracker
 
 import (
 	"hash/maphash"
+	"strconv"
 	"sync"
 	"time"
 )
@@ -65,6 +66,73 @@ func (s *expiringSet) fingerprint(key string) uint64 {
 
 func (s *expiringSet) shardFor(fp uint64) *expiringShard {
 	return &s.shards[fp%expiringSetShards]
+}
+
+// keyWriter accumulates a dedup key's components directly into a hash, so the
+// key never exists as a string. The old path built one with fmt.Sprintf per
+// webhook purely for maphash to consume and discard.
+//
+// Components are zero-separated. The concatenation this replaced was
+// ambiguous in principle — ("ab","c") and ("a","bc") produced the same key —
+// which the separator removes.
+type keyWriter struct{ h maphash.Hash }
+
+func (k *keyWriter) Str(v string) *keyWriter {
+	_, _ = k.h.WriteString(v)
+	_ = k.h.WriteByte(0)
+	return k
+}
+
+func (k *keyWriter) Int(v int64) *keyWriter {
+	var buf [20]byte
+	_, _ = k.h.Write(strconv.AppendInt(buf[:0], v, 10))
+	_ = k.h.WriteByte(0)
+	return k
+}
+
+func (k *keyWriter) Bool(v bool) *keyWriter {
+	b := byte('F')
+	if v {
+		b = 'T'
+	}
+	_ = k.h.WriteByte(b)
+	_ = k.h.WriteByte(0)
+	return k
+}
+
+// newKey returns a keyWriter seeded for this set. Returned by value and used
+// as a local so it never escapes: taking its address across a function
+// boundary (for example by passing a build callback) makes escape analysis
+// heap-allocate it, which is the allocation this whole path exists to avoid.
+//
+// maphash.Hash must not be copied after first use; this copy happens before
+// any write.
+func (s *expiringSet) newKey() keyWriter {
+	var k keyWriter
+	k.h.SetSeed(s.seed)
+	return k
+}
+
+// CheckAndAdd reports whether the key was already present and unexpired, and
+// records it with ttl when it was not.
+//
+// One hash and one lock acquisition, against the two each that a Has-miss
+// followed by Add cost. It also closes the check-then-act window in the old
+// pairing, where two workers handling the same encounter could both miss and
+// both deliver.
+func (s *expiringSet) CheckAndAdd(k *keyWriter, ttl time.Duration) bool {
+	fp := k.h.Sum64()
+	sh := s.shardFor(fp)
+	now := time.Now()
+
+	sh.mu.Lock()
+	defer sh.mu.Unlock()
+
+	if expiry, ok := sh.entries[fp]; ok && expiry > now.UnixNano() {
+		return true
+	}
+	sh.entries[fp] = now.Add(ttl).UnixNano()
+	return false
 }
 
 // Has reports whether key is present and unexpired. Entries past their expiry
--- a/processor/internal/tracker/expiring_set.go
+++ b/processor/internal/tracker/expiring_set.go
@@ -34,6 +34,7 @@ type expiringSet struct {
 	seed   maphash.Seed
 	shards [expiringSetShards]expiringShard
 	stop   chan struct{}
+	done   chan struct{}
 	once   sync.Once
 }
 
@@ -49,6 +50,7 @@ func newExpiringSet() *expiringSet {
 	s := &expiringSet{
 		seed: maphash.MakeSeed(),
 		stop: make(chan struct{}),
+		done: make(chan struct{}),
 	}
 	for i := range s.shards {
 		s.shards[i].entries = make(map[uint64]int64)
@@ -101,12 +103,20 @@ func (s *expiringSet) Len() int {
 	return n
 }
 
-// Close stops the background sweeper. Safe to call more than once.
+// Close stops the background sweeper and waits for it to exit. Safe to call
+// more than once.
+//
+// The wait matters: sweep holds a shard lock while it scans, so a Close that
+// returned early would leave DuplicateCache.Close's step of the shutdown
+// ordering non-quiescent. mute.Sweeper and snapshots.Sweeper use the same
+// stop/done/once trio.
 func (s *expiringSet) Close() {
 	s.once.Do(func() { close(s.stop) })
+	<-s.done
 }
 
 func (s *expiringSet) sweepLoop() {
+	defer close(s.done)
 	ticker := time.NewTicker(expiringSweepInterval)
 	defer ticker.Stop()
 	for {
--- a/processor/internal/tracker/expiring_set_test.go
+++ b/processor/internal/tracker/expiring_set_test.go
@@ -1,6 +1,9 @@
 package tracker
 
 import (
+	"fmt"
+	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 )
@@ -100,4 +103,110 @@ func TestExpiringSetCloseWaitsForSweeper(t *testing.T) {
 	// Close is idempotent — DuplicateCache.Close may be reached twice during
 	// a shutdown that is itself racing a signal handler.
 	s.Close()
+}
+
+func BenchmarkExpiringSetCheckAndAdd(b *testing.B) {
+	s := newExpiringSet()
+	defer s.Close()
+
+	b.ReportAllocs()
+	b.ResetTimer()
+	for i := 0; b.Loop(); i++ {
+		k := s.newKey()
+		k.Str("47281957203847502").Bool(true).Int(1500)
+		s.CheckAndAdd(&k, time.Hour)
+	}
+}
+
+func BenchmarkDuplicateCacheCheckPokemon(b *testing.B) {
+	dc := NewDuplicateCache()
+	defer dc.Close()
+	disappear := time.Now().Unix() + 3600
+
+	// Pre-generate ids so the benchmark measures the dedup path rather than
+	// strconv in the harness.
+	const n = 4096
+	ids := make([]string, n)
+	for i := range ids {
+		ids[i] = encounterIDForTest(i)
+	}
+
+	b.ReportAllocs()
+	b.ResetTimer()
+	for i := 0; b.Loop(); i++ {
+		dc.CheckPokemon(ids[i%n], true, 1500, disappear)
+	}
+}
+
+// BenchmarkExpiringSetHasThenAdd measures the pattern CheckAndAdd replaced:
+// build the key with fmt.Sprintf, hash it once for Has, then hash it again
+// for Add and take the shard lock a second time. Kept alongside its
+// replacement so the comparison is same-harness rather than across commits.
+func BenchmarkExpiringSetHasThenAdd(b *testing.B) {
+	s := newExpiringSet()
+	defer s.Close()
+
+	const id = "47281957203847502"
+
+	b.ReportAllocs()
+	b.ResetTimer()
+	for b.Loop() {
+		key := fmt.Sprintf("%s%s%d", id, "T", 1500)
+		if !s.Has(key) {
+			s.Add(key, time.Hour)
+		}
+	}
+}
+
+// TestCheckAndAddIsAtomic pins the check-then-act window that the old
+// Has-miss-then-Add pairing left open: two workers handling the same encounter
+// concurrently could both miss and both deliver. Exactly one caller must see
+// the key as new.
+func TestCheckAndAddIsAtomic(t *testing.T) {
+	s := newExpiringSet()
+	defer s.Close()
+
+	const workers = 64
+	var start sync.WaitGroup
+	var done sync.WaitGroup
+	var firsts atomic.Int64
+
+	start.Add(1)
+	for range workers {
+		done.Add(1)
+		go func() {
+			defer done.Done()
+			start.Wait()
+			k := s.newKey()
+			k.Str("same-encounter").Bool(true).Int(1500)
+			if !s.CheckAndAdd(&k, time.Hour) {
+				firsts.Add(1)
+			}
+		}()
+	}
+	start.Done()
+	done.Wait()
+
+	if got := firsts.Load(); got != 1 {
+		t.Errorf("%d workers saw the key as new, want exactly 1", got)
+	}
+}
+
+// TestKeyWriterSeparatesComponents guards against the ambiguity in the string
+// concatenation this replaced: ("ab","c") and ("a","bc") produced the same key.
+func TestKeyWriterSeparatesComponents(t *testing.T) {
+	s := newExpiringSet()
+	defer s.Close()
+
+	a := s.newKey()
+	a.Str("ab").Str("c")
+	b := s.newKey()
+	b.Str("a").Str("bc")
+
+	if s.CheckAndAdd(&a, time.Hour) {
+		t.Fatal("first key should be new")
+	}
+	if s.CheckAndAdd(&b, time.Hour) {
+		t.Error(`("a","bc") collided with ("ab","c"): components are not separated`)
+	}
 }
--- a/processor/internal/tracker/expiring_set_test.go
+++ b/processor/internal/tracker/expiring_set_test.go
@@ -8,21 +8,26 @@ import (
 // TestDuplicateCacheEntryExpires covers the behaviour the dedup path depends
 // on: once a key's TTL has passed the same webhook must be treated as fresh
 // again, whether or not the sweeper has run.
-func TestDuplicateCacheEntryExpires(t *testing.T) {
-	dc := NewDuplicateCache()
-	defer dc.Close()
+func TestExpiringSetEntryExpires(t *testing.T) {
+	s := newExpiringSet()
+	defer s.Close()
 
-	// disappear_time already in the past clamps the TTL to its 60s floor,
-	// so drive expiry through the set directly with a sub-second TTL.
-	dc.seen.Add("encounter-1", 10*time.Millisecond)
+	// The TTL has to outlast a scheduling hiccup between Add and the
+	// presence check: a GC pause or a loaded runner descheduling this
+	// goroutine would otherwise expire the entry before it is read and fail
+	// the test for reasons that have nothing to do with the code. The expiry
+	// half is safe in the other direction, since over-sleeping cannot
+	// resurrect an entry.
+	const ttl = 250 * time.Millisecond
+	s.Add("encounter-1", ttl)
 
-	if !dc.seen.Has("encounter-1") {
+	if !s.Has("encounter-1") {
 		t.Fatal("key should be present immediately after Add")
 	}
 
-	time.Sleep(30 * time.Millisecond)
+	time.Sleep(ttl + 50*time.Millisecond)
 
-	if dc.seen.Has("encounter-1") {
+	if s.Has("encounter-1") {
 		t.Error("key should read as absent once its TTL has passed, before any sweep")
 	}
 }
--- a/processor/internal/tracker/expiring_set_test.go
+++ b/processor/internal/tracker/expiring_set_test.go
@@ -81,3 +81,23 @@ func TestExpiringSetSpreadsAcrossShards(t *testing.T) {
 		t.Errorf("expected all %d shards populated, got %d", expiringSetShards, populated)
 	}
 }
+
+// TestExpiringSetCloseWaitsForSweeper asserts Close is synchronous: when it
+// returns, the sweep goroutine has actually exited and is no longer holding
+// shard locks. DuplicateCache.Close runs during the documented shutdown
+// ordering's "duplicates" step, which is only meaningfully quiescent if this
+// holds.
+func TestExpiringSetCloseWaitsForSweeper(t *testing.T) {
+	s := newExpiringSet()
+	s.Close()
+
+	select {
+	case <-s.done:
+	default:
+		t.Error("Close returned while sweepLoop was still running")
+	}
+
+	// Close is idempotent — DuplicateCache.Close may be reached twice during
+	// a shutdown that is itself racing a signal handler.
+	s.Close()
+}
--- a/processor/internal/tracker/expiring_set_test.go
+++ b/processor/internal/tracker/expiring_set_test.go
@@ -1,0 +1,78 @@
+package tracker
+
+import (
+	"testing"
+	"time"
+)
+
+// TestDuplicateCacheEntryExpires covers the behaviour the dedup path depends
+// on: once a key's TTL has passed the same webhook must be treated as fresh
+// again, whether or not the sweeper has run.
+func TestDuplicateCacheEntryExpires(t *testing.T) {
+	dc := NewDuplicateCache()
+	defer dc.Close()
+
+	// disappear_time already in the past clamps the TTL to its 60s floor,
+	// so drive expiry through the set directly with a sub-second TTL.
+	dc.seen.Add("encounter-1", 10*time.Millisecond)
+
+	if !dc.seen.Has("encounter-1") {
+		t.Fatal("key should be present immediately after Add")
+	}
+
+	time.Sleep(30 * time.Millisecond)
+
+	if dc.seen.Has("encounter-1") {
+		t.Error("key should read as absent once its TTL has passed, before any sweep")
+	}
+}
+
+// TestExpiringSetSweepReclaimsExpiredEntries asserts the sweeper actually
+// frees entries rather than only hiding them from reads — the whole point of
+// the change is bounded memory, and lazy expiry alone would never reclaim.
+func TestExpiringSetSweepReclaimsExpiredEntries(t *testing.T) {
+	s := newExpiringSet()
+	defer s.Close()
+
+	for i := range 1000 {
+		s.Add(encounterIDForTest(i), time.Hour)
+	}
+	for i := 1000; i < 2000; i++ {
+		s.Add(encounterIDForTest(i), time.Millisecond)
+	}
+
+	if got := s.Len(); got != 2000 {
+		t.Fatalf("expected 2000 entries before sweep, got %d", got)
+	}
+
+	time.Sleep(10 * time.Millisecond)
+	s.sweep(time.Now().UnixNano())
+
+	if got := s.Len(); got != 1000 {
+		t.Errorf("expected the 1000 expired entries to be reclaimed, got %d remaining", got)
+	}
+}
+
+// TestExpiringSetSpreadsAcrossShards guards the sharding: a single-shard set
+// would serialise every dedup check on one mutex at scanner throughput.
+func TestExpiringSetSpreadsAcrossShards(t *testing.T) {
+	s := newExpiringSet()
+	defer s.Close()
+
+	for i := range 10_000 {
+		s.Add(encounterIDForTest(i), time.Hour)
+	}
+
+	populated := 0
+	for i := range s.shards {
+		s.shards[i].mu.Lock()
+		if len(s.shards[i].entries) > 0 {
+			populated++
+		}
+		s.shards[i].mu.Unlock()
+	}
+
+	if populated != expiringSetShards {
+		t.Errorf("expected all %d shards populated, got %d", expiringSetShards, populated)
+	}
+}
--- a/processor/internal/tracker/rarity.go
+++ b/processor/internal/tracker/rarity.go
@@ -82,19 +82,32 @@ type StatsTracker struct {
 	groups  map[int]int        // pokemon_id -> rarity group
 	shiny   map[int]ShinyStats // pokemon_id -> shiny stats (cached)
 
-	// now is the clock, injectable so window expiry can be tested without
-	// sleeping. Defaults to time.Now().Unix.
-	now func() int64
+	// nowFunc is the clock, injectable so window expiry can be tested
+	// without sleeping. Defaults to time.Now. Set once at construction and
+	// never reassigned: recalcLoop reads it from its own goroutine, so a
+	// later write would be a data race.
+	nowFunc func() time.Time
 }
 
 // NewStatsTracker creates a new stats tracker with the given config.
 func NewStatsTracker(cfg StatsConfig) *StatsTracker {
+	return newStatsTrackerWithClock(cfg, time.Now)
+}
+
+// newStatsTrackerWithClock creates a tracker with an injectable clock, used in
+// tests to control the rolling window without wall-clock dependencies.
+// Mirrors newRateCounterWithClock in internal/webhook.
+//
+// The clock is a constructor parameter rather than an exported field because
+// recalcLoop starts here: anything assigned afterwards would be written by the
+// caller while that goroutine reads it.
+func newStatsTrackerWithClock(cfg StatsConfig, nowFunc func() time.Time) *StatsTracker {
 	st := &StatsTracker{
 		cfg:     cfg,
 		buckets: newBucketRing(cfg.WindowHours),
 		groups:  make(map[int]int),
 		shiny:   make(map[int]ShinyStats),
-		now:     func() int64 { return time.Now().Unix() },
+		nowFunc: nowFunc,
 	}
 	go st.recalcLoop()
 	return st
@@ -134,7 +147,7 @@ func (st *StatsTracker) RecordSighting(pokemonID int, ivScanned bool, isShiny bo
 	st.mu.Lock()
 	defer st.mu.Unlock()
 
-	minute := st.now() / 60
+	minute := st.nowFunc().Unix() / 60
 	b := &st.buckets[int(minute%int64(len(st.buckets)))]
 	if b.minute != minute {
 		b.reset(minute)
@@ -214,7 +227,7 @@ type counters struct {
 //
 // Caller must hold at least a read lock.
 func (st *StatsTracker) aggregate() (map[int]*counters, int64) {
-	oldest := st.now()/60 - int64(windowMinutes(st.cfg.WindowHours))
+	oldest := st.nowFunc().Unix()/60 - int64(windowMinutes(st.cfg.WindowHours))
 
 	counts := make(map[int]*counters)
 	var totalAll int64
@@ -241,7 +254,7 @@ func (st *StatsTracker) aggregate() (map[int]*counters, int64) {
 // totalInWindow counts sightings still inside the window without building the
 // per-species aggregate. Caller must hold at least a read lock.
 func (st *StatsTracker) totalInWindow() int64 {
-	oldest := st.now()/60 - int64(windowMinutes(st.cfg.WindowHours))
+	oldest := st.nowFunc().Unix()/60 - int64(windowMinutes(st.cfg.WindowHours))
 	var total int64
 	for i := range st.buckets {
 		if st.buckets[i].minute >= oldest {
--- a/processor/internal/tracker/rarity.go
+++ b/processor/internal/tracker/rarity.go
@@ -47,6 +47,14 @@ type speciesCounts struct {
 // (minute, species) instead of one record per sighting is what keeps this
 // tracker's memory bounded: size is a function of the window length and the
 // number of distinct species, never of webhook throughput.
+//
+// This ring is mechanically the same as webhook.RateCounter's: slot index is
+// (unix/60 % len), a slot whose stored minute differs is stale and reset in
+// place rather than allocated, and reads filter by age instead of sweeping.
+// The two differ in what they hold (per-species counters here, per-type counts
+// there) and in whether the window is configurable, which is why they are not
+// one type today. Keep them in step: a boundary bug found in one is almost
+// certainly present in the other.
 type statsBucket struct {
 	minute  int64 // wall-clock minute this bucket covers; -1 when unused
 	total   int64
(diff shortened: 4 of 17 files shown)


processor/internal/tracker/weather.go
```diff
@@ -47,11 +65,55 @@ type WeatherTracker struct {
...
+	wt.mu.Lock()
+	defer wt.mu.Unlock()
+
+	for cellID, cd := range wt.controllerData {
+		if cd.lastSeen < idleBefore {
+			delete(wt.controllerData, cellID)
+			continue
+		}
+		for ts := range cd.hourWeather {
+			if ts < oldestHour {
```
jfberry, line 210: **[1/15 · correctness]** Eviction can suppress genuine weather-change alerts under webhook backlog.

`evict` prunes `hourWeather` by processor wall clock, but `UpdateFromWebhook` reads the previous-hour entry keyed by the webhook's own event time. Concrete failure: a webhook stamped `updated=13:59` is processed at 14:05 after Golbat backlogs ~5 min; the 10-minute sweep (arbitrary phase) already ran past 14:00 and deleted the 12:00 entry (`ts < 14:00−3600`). `previousHourTimestamp=12:00` → `hasPrevious=false` → `changed=false`: the real 12:00→13:00 weather change emits no `WeatherChange`, so users tracking that cell get no alert and no re-evaluation of boosted pokemon. Base kept full history and would have alerted — this is a regression.

Suggested fix: keep one extra hour of slack (`weatherHistoryHours = 2`), or derive the reader lookback from the eviction constant so they can't disagree.

processor/internal/tracker/weather.go
```diff
@@ -94,6 +156,7 @@ func (wt *WeatherTracker) UpdateFromWebhook(cellID string, condition int, timest
 
 	cd.hourWeather[hourTimestamp] = condition
 	cd.lastCurrentWeatherCheck = timestamp
+	cd.lastSeen = timestamp
```
jfberry: **[2/15 · correctness]** Mixed clock domains drive cell eviction.

`cd.lastSeen` is stamped here from the webhook-supplied `Updated` field, while `evict()` compares it against local wall clock (`SetHourWeather` and `CheckWeatherOnMonster` both stamp `time.Now()`). Replaying logged webhooks (a documented workflow — `logs/webhooks.log` "can be replayed") or Golbat clock skew/backlog with `updated` >24h old creates cells instantly eligible for eviction; the next sweep wipes them, discarding previous-hour comparison state so subsequent genuine weather changes emit no alert. Conversely, a bogus future timestamp pins a dead cell forever.

Suggested fix: stamp receipt time (`time.Now().Unix()`) here, matching the other two write sites — single-clock semantics for the eviction decision.

processor/internal/tracker/weather.go
```diff
@@ -47,11 +65,55 @@ type WeatherTracker struct {
 
 // NewWeatherTracker creates a new weather tracker.
 func NewWeatherTracker() *WeatherTracker {
-	return &WeatherTracker{
+	wt := &WeatherTracker{
 		controllerData: make(map[string]*controllerCellData),
 		localData:      make(map[string]*localCellData),
 		changes:        make(chan WeatherChange, 100),
 	}
+	go wt.evictionLoop()
```
jfberry, line 152: **[4/15 · lifecycle]** `evictionLoop` has no stop channel and `WeatherTracker` has no `Close` — the only background reclaim loop in the codebase without one (`mute.Sweeper`, `snapshots.Sweeper`, the ratelimit cleanup loop, and this PR's own `expiringSet` all have stop mechanisms), and it can't participate in the documented shutdown ordering.

Every `NewWeatherTracker()` permanently leaks a goroutine + 10-minute ticker and pins the tracker's maps against GC — `weather_test.go` now constructs several per run, which blocks any future goleak adoption, and any future re-init path would accumulate unstoppable sweepers contending on `wt.mu`.

Suggested fix: mirror the `expiringSet` stop/once/Close pattern and wire it into `ProcessorService.Close`.

processor/internal/tracker/weather.go
```diff
@@ -28,14 +28,32 @@ type localCellData struct {
...
+	// weatherHistoryHours is how far back hourly entries stay reachable.
+	// UpdateFromWebhook compares against the previous hour and
+	// ExportCellWeather filters to currentHour-3600, so nothing older is
+	// readable and nothing older is kept.
+	weatherHistoryHours = 1
+
+	// weatherCellIdleSecs is how long a cell survives without a write before
+	// its state is dropped. Longer than the AccuWeather forecast refresh
+	// (8h by default) so an actively-forecast cell is never reclaimed.
+	weatherCellIdleSecs = 24 * 3600
```
jfberry, line 65: **[6/15 · config-robustness]** The comment's guarantee ("longer than the AccuWeather forecast refresh, 8h by default") only holds for the default config. `forecast_refresh_interval` is operator-settable in hours with no clamp or startup warning, and `NewWeatherTracker` takes no config so the invariant can't be checked.

An operator rationing AccuWeather quota who sets `forecast_refresh_interval` above 24 gets cells whose only writes are forecast pushes evicted between refreshes — discarding `lastCurrentWeatherCheck` and previous-hour state, so the next weather webhook finds `hasPrevious=false` and a genuine change emits no alert.

Suggested fix: either clamp/warn at config load, or pass the interval in and size the idle threshold from it.

processor/internal/tracker/weather.go
```diff
@@ -246,6 +310,8 @@ func (wt *WeatherTracker) CheckWeatherOnMonster(cellID string, lat, lon float64,
 
 	local := wt.localData[cellID]
 	controller := wt.controllerData[cellID]
+	local.lastSeen = now
```
jfberry: **[10/15 · maintainability]** Eviction liveness is hand-threaded: `lastSeen` is stamped at 3 independent write sites (`UpdateFromWebhook`, `SetHourWeather`, here), each adjacent to a byte-identical copy-pasted getOrCreate block.

Risk: a future write path (bulk forecast import, new webhook variant) copies one getOrCreate block minus the one-line `lastSeen` update; it passes every functional test — data reads back fine for 24h — then the idle sweep silently deletes cells written only by that path. Intermittent, self-healing loss of weather state; close to undiagnosable in production.

Suggested fix: one `touch()`/`getOrCreate` helper that owns the `lastSeen` stamp.

processor/internal/geo/timezone.go
```diff
@@ -10,7 +10,15 @@ var finder tzf.F
...
-	finder, err = tzf.NewFullFinder()
+	// NewDefaultFinder, not NewFullFinder: both wrap the same preindex tile
+	// fast path, but Full decodes tzf-dist's 18 MB full-precision polygon set
+	// where Default decodes the 6.5 MB topology-simplified one. That cost
+	// 140.96 MB of live heap at init in production (4.44% of the total) for
+	// accuracy we cannot use — upstream bounds the simplified boundaries to
+	// ~111 m of the true border, and a pokemon spawn that close to a timezone
+	// line resolves to a neighbouring zone with the same wall clock in every
+	// case a Poracle alert cares about.
+	finder, err = tzf.NewDefaultFinder()
```
jfberry, line 30: **[5/15 · documentation]** The justification comment overclaims. A spawn inside the ~111 m simplified-boundary band does **not** always resolve "to a neighbouring zone with the same wall clock": Nepal/India differ by 15 min, the Arizona line by 1h half the year, China's western border by up to 2.5h — and empirical comparison at the pinned tzf v1.2.5 also shows remote land→nautical-Etc flips of up to 2h (e.g. America/Punta_Arenas → Etc/GMT+5).

To be clear, the swap itself verified sound: zero empty-string returns across a 259k-point grid, and every city-center test is unaffected. But a spawn or saved user location inside the band of an offset-differing border renders despawn/hatch times and profile/summary schedules 15 min–2 h off, silently — behavior the comment asserts cannot occur.

Suggested fix: correct the comment to state the real trade-off, and optionally pin one offset-differing border point in the test table to document the accepted behavior.

processor/internal/tracker/expiring_set.go
```diff
@@ -0,0 +1,135 @@
...
+	return maphash.String(s.seed, key)
+}
+
+func (s *expiringSet) shardFor(fp uint64) *expiringShard {
+	return &s.shards[fp%expiringSetShards]
+}
+
+// Has reports whether key is present and unexpired. Entries past their expiry
+// read as absent even when the sweeper has not reclaimed them yet.
+func (s *expiringSet) Has(key string) bool {
```
jfberry, line 140: **[8/15 · efficiency]** The dedup hot path does redundant work the new design makes avoidable: every `Check*` builds a `fmt.Sprintf` key only for maphash to consume and discard, then the Has-miss→Add sequence hashes the same key twice and takes the same shard mutex twice (`Has` also calls `time.Now().UnixNano()` inside the critical section).

At scanner throughput that's ~2000–6000 transient heap allocs/sec of pure garbage plus doubled lock/hash cost per new webhook — sustained GC churn in a memory-optimization PR.

Suggested fix: a single `CheckAndAdd(ttl, components...)` that feeds `maphash.Hash` directly is zero-alloc and halves the locking — and as a free co-benefit it closes the pre-existing check-then-act window where two workers processing the same encounter concurrently both miss and deliver duplicate alerts.

processor/internal/tracker/expiring_set.go
```diff
@@ -0,0 +1,135 @@
...
+	for i := range s.shards {
+		s.shards[i].mu.Lock()
+		n += len(s.shards[i].entries)
+		s.shards[i].mu.Unlock()
+	}
+	return n
+}
+
+// Close stops the background sweeper. Safe to call more than once.
+func (s *expiringSet) Close() {
```
jfberry, line 181: **[12/15 · lifecycle]** `expiringSet` copies only half the established sweeper pattern: stop+once but no `done` channel, so `Close()` returns without waiting for `sweepLoop` to exit. `mute.Sweeper.Stop` and `snapshots.Sweeper.Stop` both `close(stop)` then block on `<-done`.

Consequence: during shutdown, `DuplicateCache.Close` → `seen.Close` can return while `sweep()` is mid-scan holding shard locks, so the documented shutdown ordering's "duplicates" step isn't actually quiescent when it returns. This is also the third copy of the identical ticker-select sweep loop (mute, snapshots, here).

Suggested fix: mirror the stop/done/once trio — or extract one shared sweeper taking a sweep func.

processor/internal/tracker/expiring_set_test.go
```diff
@@ -0,0 +1,78 @@
...
+// again, whether or not the sweeper has run.
+func TestDuplicateCacheEntryExpires(t *testing.T) {
+	dc := NewDuplicateCache()
+	defer dc.Close()
+
+	// disappear_time already in the past clamps the TTL to its 60s floor,
+	// so drive expiry through the set directly with a sub-second TTL.
+	dc.seen.Add("encounter-1", 10*time.Millisecond)
+
+	if !dc.seen.Has("encounter-1") {
```
jfberry: **[14/15 · test-flake]** The "present immediately after Add" assertion races the wall clock with a 10ms TTL — it `t.Fatal()`s if the goroutine is descheduled >10ms between Add and Has (GC pause, `-race` overhead, loaded CI runner). The expiry half is safe (over-sleeping can't hurt); only the presence half is timing-sensitive.

Suggested fix: a 250ms+ TTL keeps the test honest without slowing it. Also worth noting: the test constructs a full `DuplicateCache` but bypasses it via `dc.seen` — testing `newExpiringSet()` directly would match its sibling tests.

processor/internal/tracker/rarity.go
```diff
@@ -53,35 +75,72 @@ type ShinyStats struct {
...
+		shiny:   make(map[int]ShinyStats),
+		now:     func() int64 { return time.Now().Unix() },
 	}
 	go st.recalcLoop()
 	return st
 }
 
+// newBucketRing sizes the ring to cover the whole window inclusively: one
+// bucket per minute plus the current partial minute.
+func newBucketRing(windowHours int) []statsBucket {
```
jfberry, line 126: **[9/15 · reuse]** This per-minute bucket ring (slot = `unix/60 % N`, stale-slot detection via stored minute, in-place `clear(map)` reset, lazy age filtering at read time) re-implements the exact mechanism of `webhook.RateCounter` (`internal/webhook/rate.go`). Two hand-rolled minute rings with the same subtle invariants (wrap recycling, stale-slot skip, window sizing) now have to be kept correct independently — a boundary bug fixed in one must be re-found in the other.

Related consistency nits: the injectable clock here is `now func() int64`, diverging from the `nowFunc func() time.Time` convention used by RateCounter, delivery/ratelimit, telegram, and the schedulers (a fifth clock-injection shape hides from a `nowFunc` grep); and `totalInWindow` copy-pastes `aggregate`'s window-cutoff expression — an `oldestMinute()` helper would keep the two window definitions from drifting.

Suggested fix: extract or explicitly mirror the RateCounter core.

processor/internal/tracker/rarity.go
```diff
@@ -53,35 +75,72 @@ type ShinyStats struct {
...
-	st.mu.Unlock()
+	defer st.mu.Unlock()
+
+	minute := st.now() / 60
+	b := &st.buckets[int(minute%int64(len(st.buckets)))]
+	if b.minute != minute {
+		b.reset(minute)
+	}
+
+	c := b.species[int32(pokemonID)]
```
jfberry, line 164: **[15/15 · robustness]** Unchecked `int32()` narrowing of `pokemonID`. The webhook receiver is unauthenticated and doesn't range-check `pokemon_id` before this call: a garbage ID like 2³¹+25 becomes `int32 -2147483623` and round-trips through `aggregate()` as a nonsense species key, polluting rarity/shiny exports. Real game IDs are 4 digits so the impact is stats pollution only — a one-line bounds guard removes it.

jfberry (commented):
## Review (high-effort pass)

First, the headline claims check out. All three swaps were verified as drop-in equivalent:

- **rarity slice → bucketed minute ring**: window/wrap-around arithmetic traced correct, including slot recycling
- **dedup ttlcache → sharded 64-bit fingerprint set**: expiry semantics match; the collision math is honest (~one dropped alert per ~49 years at stated throughput)
- **tzf `NewFullFinder` → `NewDefaultFinder`**: compared empirically on a 259k-point grid — zero empty-string returns, all existing city-center tests unaffected

PR head builds, vets, lints clean and passes `go test` with and without `-race`. Nice work — the RSS win looks real.

15 findings, ranked by severity: 13 inline below, plus these 2 that fall outside the diff hunks:

**[3/15 · correctness] `processor/internal/tracker/rarity.go:169` — `refresh_interval_mins = 0` panics the processor at startup.** `recalcLoop` calls `time.NewTicker(RefreshIntervalMins * time.Minute)` with no floor. An explicit `refresh_interval_mins = 0` in `[stats]` survives config defaulting (toml.Unmarshal decodes over the pre-populated defaults struct) and hits an unrecoverable `non-positive interval for NewTicker` panic at boot. The line itself is pre-existing, but this PR rewrote the file around it and already added exactly the needed zero-floor to the sibling field in `windowMinutes()` — mirroring that one-liner here closes it.

**[11/15 · memory-leak] `processor/internal/tracker/accuweather.go:39` — the stranded-cell leak this PR targets persists in the sibling component.** `AccuWeatherClient`'s `cellMutexes`/`cellLocations`/`cellForecasts` share the same cell keyspace but are never evicted (no `delete()` anywhere), and `WeatherTracker.evict` doesn't notify it. On a forecast-enabled install, shifting the scan area reclaims WeatherTracker cells after 24h but leaves one mutex + location + forecastState per abandoned S2 cell forever — the same slow growth pattern this PR set out to eliminate, one struct over.

**Verdict**: nothing here is a merge-blocker for a typical deployment, but findings 1–3 (the two weather-eviction clock issues and the ticker panic) are worth fixing before merge — the first two silently suppress genuine weather-change alerts under Golbat backlog, which is a behavioral regression vs base.

🤖 Review assisted by [Claude Code](https://claude.com/claude-code)

processor/internal/tracker/rarity_test.go
```diff
@@ -112,3 +113,99 @@ func TestStatsTrackerReset(t *testing.T) {
...
+	const limit = 30 << 20
+	if growth > limit {
+		t.Errorf("recording %d sightings grew the heap by %.1f MB, want <= %d MB — memory is scaling with sighting count",
+			sightings, float64(growth)/(1<<20), limit>>20)
+	}
+}
+
+// TestStatsTrackerDropsSightingsOutsideWindow covers behaviour the slice-based
+// tracker had but never tested: sightings older than window_hours stop counting.
+func TestStatsTrackerDropsSightingsOutsideWindow(t *testing.T) {
```
jfberry, line 168: **[7/15 · test-contract]** `TestStatsTrackerDropsSightingsOutsideWindow` only passes because `testStatsConfig` sets `MinSampleSize: 0`. With any production `min_sample_size` (default 10000), `recalculate()` skips the groups update when the window drains (`totalAll < min`), so stale rarity groups persist and `GetRarityGroup` does **not** return `RarityUnknown` as the test name promises.

The retention behavior is intentional (inherited from base), but the test codifies "sightings older than window_hours stop counting" as if unconditional — a future refactor trusting it as the spec, or an operator debugging stale rarity after an overnight scanner outage, is misled while this stays green.

Suggested fix: add a `MinSampleSize > 0` companion test pinning the actual behavior, or a comment making the contract explicit.

processor/internal/tracker/rarity_test.go
```diff
@@ -112,3 +113,99 @@ func TestStatsTrackerReset(t *testing.T) {
...
+	cfg := testStatsConfig()
+	cfg.WindowHours = 1
+	st := NewStatsTracker(cfg)
+
+	now := int64(1_700_000_000)
+	st.now = func() int64 { return now }
```
jfberry: **[13/15 · test-race]** Tests assign `st.now` (and mutate the captured `now` variable) after `NewStatsTracker` has already started the `recalcLoop` goroutine, with no synchronization — a latent data race that stays quiet only because `testStatsConfig`'s 60-min refresh ticker never fires within test lifetime. Anyone shrinking `RefreshIntervalMins` in `testStatsConfig` (natural for a future recalc test) turns this into a `-race` failure or a torn read of the captured int64.

Suggested fix: accept the clock via the constructor, or set it under `st.mu.Lock()`.

TurtIeSocks:
Thanks — this caught two things I had genuinely got wrong, not just untidy.

All 15 findings are addressed across 15 commits, one per finding. Gate is green
(`go build`, `go vet`, `go test -count=1 ./...`, `golangci-lint run` at 0 issues),
and the package also passes with `-race`.

## The three you flagged for merge

**1/15 — history vs event-time reads.** Confirmed and fixed. Your backlog
scenario reproduces exactly: `TestWeatherTrackerKeepsHistoryForBackloggedWebhooks`
stamps a webhook late in hour N, sweeps at N+1, and gets no `WeatherChange`
without the fix. Retention is now two hours, documented as one hour of reader
lookback plus one hour of slack for event-time lag, with a note not to lower it
to match the reader literally. The existing retention test moved to the new
contract.

**2/15 — mixed clock domains.** Fixed as suggested: liveness stamps receipt
time. `WeatherTracker` now holds an injectable `nowFunc` and every liveness read
and write goes through it, including the read paths deriving `currentHour`.
Without the injectable clock the case can only be tested by backdating a
webhook, which is the exact conflation being removed.
`TestWeatherTrackerReplayedWebhookIsNotInstantlyIdle` covers the replay path.
Worth noting my original `TestWeatherTrackerEvictsIdleCells` passed only by
relying on the buggy behaviour, so it is rewritten to advance the clock instead.

**3/15 — ticker panic.** Fixed, mirroring `windowMinutes` as you suggested. The
computation moved into `recalcInterval` because a panic in a spawned goroutine
cannot be recovered by the test that triggers it, so asserting on the duration
is the only way to pin it.

## Two I checked rather than took on trust

**5/15 — the tzf comment.** You were right and the comment was wrong. I scanned
the India/Nepal border comparing both finders and found ~65 points inside the
band where the offsets differ, not just the zone names. The comment now states
the real trade, and `TestGetTimezone_SimplifiedBoundaryOffsets` pins (26.5000,
88.1000) where `NewFullFinder` gives `Asia/Kathmandu` and the finder we ship
gives `Asia/Kolkata`, so a tzf data update that moves the boundary shows up as a
test change.

**8/15 — the dedup hot path.** Your zero-alloc claim holds, but not with the
obvious shape: passing `&keyWriter` to a build callback makes escape analysis
heap-allocate it, which measured 1 alloc/op. Returning the hasher by value and
keeping it local gets to 0. Same-harness benchmarks are in the tree:

```
Has-then-Add   77.97 ns/op   24 B/op   1 allocs/op
CheckAndAdd    54.03 ns/op    0 B/op   0 allocs/op
```

`CheckPokemon` goes 222.5 ns/op / 5 allocs to 84.10 ns/op / 0. The
check-then-act co-benefit you predicted is real and now pinned by
`TestCheckAndAddIsAtomic`: exactly one of 64 racing callers sees the key as new.
Components are zero-separated too, which removes an ambiguity in the old
concatenation where `("ab","c")` and `("a","bc")` produced the same key. Key
identity and every TTL are otherwise unchanged.

## 9/15 — mirrored rather than extracted

Taking the two concrete parts: the duplicated window cutoff is now a single
`oldestMinute()`, and the clock shape moved to `nowFunc func() time.Time` under
7/15 so all clock injection greps under one name.

On the ring itself I took the "explicitly mirror" branch. Both types now
document the shared invariants and point at each other, so a boundary bug found
in one gets looked for in the other. I stopped short of extracting a shared
type because it means refactoring `RateCounter` and its tests inside a branch
that is already large and currently running in production. Say the word and I
will do it as its own PR, where it can be reviewed on its own terms.

## Everything else

- **4/15** `WeatherTracker` gets the same stop/done/once trio as the sibling
  sweepers, wired into `ProcessorService.Close` after the webhook and render
  workers are quiescent.
- **6/15** `WithForecastRefreshInterval` sizes the idle window to two refresh
  periods when the configured cadence would outlast the default, so the
  invariant holds by construction rather than by the operator staying under an
  undocumented ceiling. Two periods so a single delayed push does not strand a
  cell either.
- **7/15** Added `TestStatsTrackerKeepsStaleGroupsBelowMinSampleSize` pinning
  the real behaviour, and noted the `MinSampleSize: 0` dependency in the sibling
  test so the two are read together.
- **10/15** `touchController` / `touchLocal` own both the create-if-missing and
  the stamp. `TestWeatherTrackerEveryWritePathStampsLiveness` drives all three
  public write paths and fails if either helper drops the stamp.
- **11/15** `evict` now collects the cell ids it drops and hands them to an
  `onEvict` callback, wired to `AccuWeatherClient.ForgetCells`. Fired outside
  `wt.mu` so a callback reaching back cannot deadlock the sweep. Putting the
  notification on the tracker keeps one owner of the liveness decision, since it
  already holds the clock, the idle window and the forecast sizing.
- **12/15** stop/done/once, matching `mute.Sweeper` and `snapshots.Sweeper`.
- **13/15** Clock is a constructor parameter via `newStatsTrackerWithClock`,
  mirroring `newRateCounterWithClock`. Constructor injection alone would not
  have been enough, since a plain captured variable is racy however it arrives,
  so the test clock is backed by `atomic.Int64`.
- **14/15** 250ms TTL, and the test constructs `newExpiringSet` directly instead
  of reaching through `DuplicateCache.seen`.
- **15/15** Non-positive and above-`MaxInt32` ids are dropped before narrowing.

Every new test was written before its fix and verified to fail against the old
code, or against a deliberately broken version where it guards pre-existing
behaviour.

processor/internal/tracker/weather_test.go
```diff
@@ -56,3 +57,229 @@ func TestWeatherTrackerInference(t *testing.T) {
...
+// TestWeatherTrackerEvictsIdleCells asserts whole cells are reclaimed once
+// they stop being scanned, so a shifted scan area does not strand them.
+//
+// Idleness is measured from when we last *received* something for the cell,
+// so the clock is driven forward here rather than backdating the webhook.
+func TestWeatherTrackerEvictsIdleCells(t *testing.T) {
+	wt := NewWeatherTracker()
+
+	clock := time.Unix(1_700_000_000, 0)
+	wt.nowFunc = func() time.Time { return clock }
```
jfberry: **[N1 · test-race]** These new tests reassign `wt.nowFunc` after `NewWeatherTracker` has already started `evictionLoop` (also lines 143, 236, 273 and `accuweather_evict_test.go:16`) — the exact latent unsynchronized-write race `ddc5b770` just eliminated for `StatsTracker` by moving the clock into the constructor. `evictionLoop` reads `wt.nowFunc` from its own goroutine; the tests write the field (and mutate the captured clock variable) with no synchronization. It stays quiet only because the 10-minute ticker can't fire within a test's lifetime — anyone shortening `weatherEvictInterval` or adding a fast-tick test turns this into a `-race` failure.

Suggested fix: add a `WithClock`/`WithNowFunc` `WeatherOption` (the options plumbing now exists via `WithForecastRefreshInterval`) and reuse the race-free `testClock`, as `StatsTracker` now does.

processor/internal/tracker/weather.go
```diff
@@ -28,32 +29,202 @@ type localCellData struct {
...
+			}
+		}
+	}
+
+	for cellID, ld := range wt.localData {
+		if ld.lastSeen < idleBefore {
+			delete(wt.localData, cellID)
+			if _, stillControlled := wt.controllerData[cellID]; !stillControlled {
+				// Already reported above if it had controller state too.
+				if !slices.Contains(dropped, cellID) {
```
jfberry: **[N3 · efficiency]** `slices.Contains(dropped, cellID)` inside the `localData` loop while holding `wt.mu` is O(n²) in the number of cells dropped by one sweep. A shifted scan area on a large install can drop tens of thousands of cells in a single sweep — ~10k dropped cells means ~50M string comparisons under the write lock, stalling every webhook path for that burst. One-time event and merely a stall, but a `map[string]struct{}` membership set makes it linear.

jfberry (commented):
## Re-review at `e5e20dc7` — all 15 findings resolved ✅

Empirical verification at the new head: `go build`, `go vet`, `go test -race -count=1` (tracker, geo, webhook), and `golangci-lint` all pass — including the newly pinned India/Nepal border timezone test. Thanks for the thorough turnaround; several fixes are nicer than what the review suggested (sizing the idle window to 2× the configured forecast refresh instead of clamping; the `keyWriter` separator injection also fixing a pre-existing key-concat ambiguity).

| # | Finding | Status |
|---|---------|--------|
| 1 | Weather eviction vs backlogged webhooks | **Fixed** — `949aab1a`: `weatherHistoryHours = 2` + regression test; >1h backlog loss is now an explicitly documented trade |
| 2 | `lastSeen` webhook-time vs wall-clock evict | **Fixed** — `01d1bdc5`: receipt-time stamping in touch helpers + replay test |
| 3 | `NewTicker(0)` boot panic | **Fixed** — `3a1e7786`: `recalcInterval()` floors to default |
| 4 | Unstoppable `evictionLoop` | **Fixed** — `1d6cd8a8`: stop/done/once + `Close()`, wired into `ProcessorService.Close` |
| 5 | tzf comment overclaim | **Fixed** — `81ec3550`: honest offsets + pinned boundary test |
| 6 | 24h idle vs unclamped forecast refresh | **Fixed** — `b9f16d6d`: idle window sized 2× configured refresh |
| 7 | `MinSampleSize: 0` masking test contract | **Fixed** — `a6ca115c`: caveat + companion test pinning stale-persistence |
| 8 | Dedup double hash/lock + TOCTOU | **Fixed** — `73ac4ecf`: zero-alloc `keyWriter` + atomic `CheckAndAdd` + benchmarks |
| 9 | Minute ring duplicates `RateCounter` | **Addressed** — `e5e20dc7`/`ddc5b770`: documented mirroring, `nowFunc` convention, `oldestMinute()`; reasonable resolution |
| 10 | Hand-threaded `lastSeen` | **Fixed** — `ac4a5fc7`: `touchController`/`touchLocal` + invariant test |
| 11 | AccuWeather cell maps never evicted | **Fixed** — `c4f9cda7`: `SetOnEvict` → `ForgetCells`, callback outside `wt.mu` |
| 12 | `expiringSet.Close` no wait | **Fixed** — `08f73df0`: done channel + idempotency test |
| 13 | `st.now` post-start assignment race | **Fixed** — `ddc5b770`: constructor clock + race-free `testClock` (but see new note below) |
| 14 | 10ms-TTL flaky assertion | **Fixed** — `281b47f5`: 250ms TTL, targets the set directly |
| 15 | `int32(pokemonID)` truncation | **Fixed** — `46afde60`: range guard + leak test |

**Verdict: mergeable as-is.** Three minor new items in the delta — two inline below, plus this one that anchors to unchanged lines:

**[N2 · test-hygiene] `weather_test.go:29,41,66,111,140,163` — six tests construct `NewWeatherTracker` without `Close`,** leaking the eviction goroutine + ticker per test — the leak `Close` was added specifically to prevent (the newer tests at 233/270 and both accuweather tests do `defer wt.Close()`). Adding the six `defer wt.Close()` calls would let this package adopt goleak.

None of the three is a blocker; happy for them to land as a follow-up.

One doc nit for the maintainer rather than this PR: CLAUDE.md's shutdown-ordering list doesn't yet mention the new weather-tracker step.

🤖 Review assisted by [Claude Code](https://claude.com/claude-code)

TurtIeSocks:
All three delta items fixed, plus the doc nit. Gate green including `-race`.

**N1 — the weather test clock race.** Fair hit, and an embarrassing one: this is
the same race `ddc5b770` removed from `StatsTracker` earlier in the branch,
reintroduced one file over. Fixed the way you suggested, with `WithClock`
joining the options plumbing so the clock is set during construction and never
reassigned, and the tests moved to the atomic-backed `testClock` the stats tests
already use. Constructor injection on its own would not have been enough, since
a plain captured variable stays racy however it reaches the tracker.

**N2 — leaked trackers in tests.** Five sites needed it rather than six: the
`WithClock` migration in N1 already paired two of them with a `Close`. Every
`NewWeatherTracker` in the package now has a matching `defer wt.Close()`, so
goleak is adoptable whenever you want it.

**N3 — quadratic dedup under the write lock.** Confirmed and worse than I
expected. Benchmarked the scan-area-shift sweep at 10k cells:

```
linear scan   118,783,117 ns/op
set             1,208,492 ns/op
```

118 ms of stall on every webhook path, 98x. Membership moved to a map and the
slice is built once at the end.
`TestWeatherTrackerEvictReportsEachCellOnce` pins that a cell present in both
`controllerData` and `localData` is still reported exactly once, and
`BenchmarkWeatherEvictDropsManyCells` keeps the scaling honest.

**Doc nit.** Taken here rather than left for you, since the branch is what made
it stale: `CLAUDE.md`'s shutdown ordering now lists the weather tracker between
duplicates and the rate limiter, verified against `ProcessorService.Close`.

One thing worth flagging on 1/15 rather than leaving it implicit in the commit
message: `weatherHistoryHours = 2` buys one hour of event-time slack, so a
Golbat backlog longer than an hour still drops the previous-hour comparison and
swallows that change. Two hours covers the backlog I can actually justify a
number for. If you would rather it were derived from something real, the honest
input is whatever bound you would put on Golbat lag in the worst case you have
seen, and I would rather take that from you than pick a bigger constant and
imply it is safe.

jfberry:
Verified the latest push at `93c349ea` — all three follow-up items are resolved (WithClock option + race-free test clock, all trackers now closed in tests, evict dedup swapped to a set with an exactly-once test and benchmark), and the full gate is green: build, vet, `go test -race`, golangci-lint, 0 issues. Thanks also for picking up the CLAUDE.md shutdown-ordering note.

That's 15/15 original findings plus 3/3 follow-ups — we're happy with this and ready to merge. 🎉

@TurtIeSocks one last check before we do: can you confirm your work here is done (nothing else you're planning to push to this branch)? Once you confirm, we'll merge.

🤖 Review assisted by [Claude Code](https://claude.com/claude-code)

TurtIeSocks:
All set here. Numbers in prod are looking good now at about 650-700MB after 2 hours.

Mygod:
The patch introduces incorrect identifiers and a destructive failure path in the v2 tracking API, plus a race that can panic forecast-enabled processors. It also loses custom-template state across restarts.

Full review comments:

- [P1] Copy generated UIDs into POST response rows — processor/internal/api/v2_tracking.go:411-413
  When this endpoint uses the SQL stores, `ApplyDiff` discards each `Insert` return value, and unlike `MockTrackingStore`, `sqlTrackingStore.Insert` does not mutate the row. New rules are therefore returned with `uid: 0`, while updated rules retain the deleted UID rather than the replacement UID, so clients cannot address the returned rule with GET, PUT, or DELETE.

- [P1] Keep in-flight forecast state alive during eviction — processor/internal/tracker/accuweather.go:92-94
  When the eviction sweep races with `EnsureForecast` after it has acquired a cell mutex or started its HTTP request, these deletes remove state that the request later dereferences without a nil check. This can panic the processor, and deleting the mutex also permits a second same-cell request to run concurrently; reclamation must be synchronized with the per-cell request lifecycle.

- [P1] Make full replacement atomic before deleting the rule — processor/internal/api/v2_tracking.go:440-444
  If the insert fails after this delete, such as during a transient MySQL error, PUT returns 500 after permanently removing the existing tracking rule; retrying the same URI then returns 404. The replacement should use a transaction or otherwise avoid committing the deletion until the insertion succeeds.

- [P2] Restore the tracked template when loading cache entries — processor/internal/delivery/tracker.go:42-42
  After a restart, `Load` reconstructs active `TrackedMessage` values without copying this newly persisted field. A subsequent prior-only `monsterChanged` event reads an empty `prior.Template` and renders the default template instead of the custom template used by the original alert.

TurtIeSocks:
Thanks @Mygod — I verified all four against the code rather than taking them at
face value. All four are real. One is this branch's fault and is fixed in
`9fe865ff`; the other three pre-date it and I'd rather not bolt them onto an
approved PR.

## The forecast eviction race — mine, fixed

Confirmed, and the mechanism is exactly as you described. `c4f9cda7` (this
branch) added the first code path that ever deletes from `cellForecasts` and
`cellMutexes`. The two re-reads you point at were safe only because nothing
could delete underneath them:

- `EnsureForecast` releases `aw.mu`, blocks on the per-cell mutex, then
  re-reads `cellForecasts[cellID]` and dereferences with no nil check.
- `fetchForecast` does the same after its HTTP call, which can run to the
  client's 15s timeout.

I reproduced it rather than reasoning about it: reverting the guard turns the
new `TestFetchForecastTailToleratesEvictedCell` into
`panic: runtime error: invalid memory address or nil pointer dereference`.

Your second point stands too, and I had missed it. An in-flight request holds
the old mutex *object*, so once its map slot is deleted a second caller for the
same cell allocates a fresh mutex, locks it uncontended, and runs a concurrent
fetch — defeating the serialization the mutex exists for and spending a second
call against the quota.

Fixed at the source rather than by nil-checking each dereference: eviction now
respects the request lifecycle. `EnsureForecast` takes an in-flight hold before
releasing `aw.mu`, and `ForgetCells` skips any held cell, reclaiming it on the
next sweep once the request completes. The nil checks stay as belt and braces,
and `fetchForecast`'s write-back moved into `storeForecastTimeout` so the guard
lives with the state it protects. `TestForgetCellsSpareInFlightRequests` covers
both halves and fails against a sabotaged guard.

Worth stating plainly: this was a regression I introduced. I added a deletion
path without auditing who else held pointers into those maps.

## The other three — real, but not from this branch

`git diff origin/develop..HEAD` touches no files under `internal/api/` or
`internal/delivery/` at all, so none of these three are reachable from this
diff. I verified each anyway:

**POST returns `uid: 0`** — true and production-reachable. `ApplyDiff` discards
`Insert`'s return value for both inserts and updates, `sqlTrackingStore.Insert`
does not mutate the row (only `MockTrackingStore` does), and
`buildRuleEnvelopes` reads `UID` straight off the row. Production wires the SQL
store, so POST responses really do carry an unaddressable uid. Fix is to capture
the returned uid and `setUID` in both `ApplyDiff` loops, mirroring what
`v2HandlePut` already does on the single-row path.

**PUT delete-then-insert** — true as described. Both calls go straight to the
bare `*sqlx.DB` with no `Tx`, so a failed insert leaves the row permanently
deleted and a retry 404s. Needs a transactional `ReplaceByUID` (or `Tx`-scoped
store variants) rather than a local tweak, which is the main reason I'd rather
it not ride along here.

**`Load` drops `Template`** — true. `Save` persists it (whole-struct copy), but
`Load` reconstructs `TrackedMessage` field by field and omits it, so a restart
silently downgrades `monsterChanged` follow-ups to the default template.
`git log -S 'Template' -- internal/delivery/tracker.go` points at `f17f3972`,
which added the field and updated the render path but not `Load`'s literal. The
one-line fix is `Template: entry.Message.Template`, though copying the struct
wholesale would stop the next added field repeating it.

Happy to open a separate PR for those three — say the word and I'll do it once
this merges, or leave them to you if you'd rather. Keeping them off this branch
is about review surface, not about doubting the findings.

Mygod:
the word

Mygod:
The in-flight hold prevents the reported panic and duplicate request, but skipped evictions are not guaranteed to be retried. Failure paths can therefore reintroduce permanent per-cell state growth.

Review comment:

- [P2] Reclaim cells skipped during an in-flight request — processor/internal/tracker/accuweather.go:98-103
  When the sweep hits a cell whose request later exits before `SetHourWeather`—for example, quota exhaustion, HTTP/decode failure, or no usable forecast—this `continue` preserves its state after `WeatherTracker.evict` has already removed the cell. Since `onEvict` reports a cell only when deleting it, no later sweep retries `ForgetCells`; if the abandoned cell is not reactivated, its mutex, location, and forecast state leak indefinitely. Record a pending eviction and reclaim it when `doneInFlight` releases the final hold.

TurtIeSocks:
Confirmed and fixed in `ce39fea2`. You're right, and the previous commit message
was wrong where it claimed the next sweep would pick these up.

Three things had to hold for the leak, and all three do:

- `WeatherTracker.evict` adds to its dropped set only inside the branch that
  deletes the cell, so `onEvict` names a cell exactly once, as it leaves the
  tracker.
- `ForgetCells`' `continue` skipped the reclaim outright.
- Those three deletes were the only ones for these maps anywhere in the file.

So a refused cell was never offered again, and if the request then exited
without reactivating it, its mutex, location key and forecast state stayed for
the life of the process.

`ForgetCells` now records the refusal in `pendingEvict` and `doneInFlight`
completes it when the last hold releases, which is the shape you suggested. The
three deletes moved into a `forget()` helper so the deferred and immediate paths
cannot drift apart.

One deliberate trade, noted in the code: a cell reactivated while its request
was running gets reclaimed too, costing one extra location lookup the next time
it is needed. Avoiding that would mean reading tracker state under `aw.mu`, and
`EnsureForecast` already goes `aw.mu` then `wt.mu` via `hasHourWeather`, so the
alternative is a lock-ordering hazard rather than a free improvement. Reclaiming
one cell early seemed the better side to err on.

`TestForgetCellsReclaimsAfterInFlightRequestCompletes` covers the
abandoned-request path and fails without the change;
`TestForgetCellsDoesNotReclaimUnevictedCells` pins that a request completing on
a cell nobody asked to evict leaves it alone. Gate green including `-race`.

Mygod:
Deferred eviction can discard newly refreshed state and unnecessarily consume additional AccuWeather quota. Focused Go tests could not run because required modules were absent and the read-only module cache could not download them.

Review comment:

- [P2] Preserve fresh state when a deferred eviction becomes stale — processor/internal/tracker/accuweather.go:148-149
  If the sweep overlaps a successful `EnsureForecast`, `fetchForecast` re-creates the tracker cell with `SetHourWeather` and records a fresh `forecastTimeout` before this defer runs. Unconditionally calling `forget` then discards that timeout and location key even though the cell is active; the next alert before the configured refresh sees next-hour data but a zero timeout, so it performs another location lookup and 12-hour forecast, charging two extra requests against the default 500/day quota. Clear the pending eviction when the fetch repopulates the cell, or verify its eviction generation before reclaiming it.

TurtIeSocks:
Valid, and fixed in `f5a44d9f`.

Worth being direct about this one: I had already spotted the behaviour and
shipped it anyway, calling it a deliberate trade in `ce39fea2`'s comment and in
my last reply. The justification I gave — that avoiding it would mean taking
`wt.mu` under `aw.mu` — was wrong, and I asserted it instead of checking it.

`EnsureForecast` already takes `aw.mu` and then `wt.mu` via `hasHourWeather`
(accuweather.go, first lock then the `hasNextHour` call), and `evict` releases
`wt.mu` before invoking `onEvict` (weather.go:231 captures it, :237 calls it
after the unlock). There is no `wt.mu` to `aw.mu` edge anywhere, so consulting
the tracker at reclaim time is the same order that already ships. I invented a
constraint and used it to justify the worse design.

`doneInFlight` now re-checks the tracker before completing a deferred eviction
and skips a cell that came back, whether it was `SetHourWeather` from a
successful fetch or an ordinary webhook that reactivated it. That keeps the
location key and the forecast timeout, so the next alert does not re-pay for a
lookup plus a 12-hour forecast.

Both directions are pinned now rather than just the one I happened to be
thinking about: `TestDeferredEvictionSkippedWhenCellCameBack` fails if the
tracker check is removed, and last round's
`TestForgetCellsReclaimsAfterInFlightRequestCompletes` fails if cells are
always treated as live. The pair brackets the behaviour instead of guarding one
edge of it.

Noted that you couldn't run the tests — read-only module cache. For what it is
worth, all three of your findings against this branch have now been confirmed
by reproduction on my side, this one included, so the missing test run doesn't
seem to have cost you anything.

Gate green: build, vet, `go test -count=1 ./...`, `-race` on the tracker
package, golangci-lint 0 issues.

jfberry:
@TurtIeSocks re: the three pre-existing findings — no separate PR needed from you, you can drop all three from your plate:

- **A** (v2 POST `uid: 0`) and **C** (`Load` dropping `Template`) are fixed in #189, with regression tests — including a SQL-faithful mock for the ApplyDiff test so the `MockTrackingStore` row-mutation can't mask that class of bug again.
- **B** (PUT delete-then-insert atomicity) we'll own here. For context: the processor currently uses no DB transactions anywhere, so a proper fix is a transactional `ReplaceByUID` across the store abstraction — a deliberate piece of design work rather than a bolt-on, and the failure window is narrow and client-recoverable in the meantime.

Thanks for verifying all four so carefully rather than taking them at face value — and to @Mygod for the sharp catches. This branch just needs the one open item (the deferred-eviction staleness note above) and then we'll merge.

Mygod:
The patch contains a non-atomic replacement path that can delete tracking data and several user-visible regressions involving persistence, API operations, mutes, templates, and tracking validation. These issues should be addressed before the change is considered correct.

Full review comments:

- [P1] Make tracking replacement atomic — processor/internal/api/v2_tracking.go:439-445
  When a PUT translates to a row whose unique key already exists, such as an invasion row with the same `(id, profile_no, gender, grunt_type)` as another rule, the delete succeeds and the subsequent insert fails with a duplicate-key error, returning 500 after permanently removing the original rule. A transient insert failure has the same data-loss behavior; perform the replacement in one database transaction.

- [P2] Restore templates when loading tracked messages — processor/internal/delivery/tracker.go:42-42
  When an active reply-tracked alert is saved and the processor restarts, persistence serializes this field but `Load` reconstructs `TrackedMessage` without copying it. Pokemon change fanout reads `prior.Template`, so a prior-only recipient receives the default `monsterChanged` template after restart instead of the source rule's template; restore `entry.Message.Template` during loading.

- [P2] Notify eviction only after the whole weather cell is gone — processor/internal/tracker/weather.go:219-219
  When local inference ages out but controller data remains fresh through weather webhooks or `SetHourWeather`, this still adds the cell to the eviction callback set. The callback invokes `AccuWeatherClient.ForgetCells`, discarding location and forecast state for an active cell and causing avoidable API fetches on its next alert; report the cell only when neither weather map retains it.

- [P2] Treat unchanged location updates as successful — processor/internal/store/user_locations_sql.go:61-61
  For a PUT that supplies the existing location's current coordinates, MySQL normally reports zero affected rows because no values changed. The store consequently returns `ErrLocationNotFound` and the endpoint responds with 404 even though the row exists; distinguish matched rows from changed rows or treat this no-op update as successful.

- [P2] Canonicalize numeric mute values before storing them — processor/internal/api/v2_mutes.go:128-128
  For Pokemon and tracking scopes, a value such as `025` passes validation but is stored verbatim. Scope matching compares it with the canonical decimal string `25`, so the successfully created mute never suppresses that event; return `strconv.Itoa(n)` after validating numeric values.

- [P2] Allow deleting platform-agnostic templates — processor/internal/api/huma_dts_reads.go:145-146
  Saving an override for a platform-agnostic type such as `help` intentionally accepts an empty platform, but this validation rejects the same empty platform before deletion reaches `DeleteEntry`. Such a valid override therefore cannot be removed through the API; apply the same platform-agnostic exemption used by the save endpoint.

- [P2] Count costume as a filter for catch-all tracking — processor/internal/bot/commands/track.go:75-75
  For non-admin users, `!track everything costume:<valid costume>` is rejected by the bare-everything guard because its `hasFilters` calculation omits this new costume value. Since matching supports Pokemon ID 0 constrained to a specific costume, count `costume != 9000` as a filter.

- [P2] Enforce the PVP league value set — processor/internal/api/v2_pokemon.go:58-58
  Although `pvp_ranking_league` documents a closed set, arbitrary integers are accepted and persisted. A value such as 42 indexes the rule under a league the PVP matcher never traverses, so the API reports success for a rule that can never alert; validate the field against 0, 500, 1500, and 2500.

jfberry:
Follow-up on the latest batch: we verified all eight against the code — every one is real. Ownership split:

**Item 2 (tracker `Load` dropping `Template`)** — already fixed: #189 (merged to develop) replaces `Load`'s field-by-field rebuild with a wholesale struct copy, so the next field added to `TrackedMessage` can't repeat the drop.

**Item 3 (`weather.go:219`)** — @TurtIeSocks this is the one item that belongs to this branch: a regression from `d5df1997` — the old `stillControlled` guard's don't-report-live-cells semantics were lost when the dedup moved to a set (our re-review endorsed that hunk and missed it too). One-liner: add local-loop cells to the dropped set only when `controllerData` no longer holds the cell. That's the last blocker; everything else below is pre-existing develop and stays off this PR.

**Item 1 (PUT atomicity)** — confirmed, and sharper than our earlier framing: `invasion` and `lures` carry legacy UNIQUE keys, so a PUT whose body collides with another rule's identity is *deterministic* data loss, not a transient window. Correction to our earlier comment: the fix won't be transactions. This schema deliberately manages integrity in application code — the FK constraints were stripped for exactly that reason, and these two unique keys are leftovers the removal missed. We'll finish the job: a migration drops the two leftover tracking keys, and v2 PUT gains the same app-level identity validation the insert path already performs (collision with a different rule → 409 before anything is deleted) across all tracking types — which also closes the silent-duplicate gap the eight keyless tables have on PUT today.

**Items 4–8** — all confirmed pre-existing; fixes PR from us shortly. One note on item 6: it needs a second half beyond the validation asymmetry — even with the platform check relaxed, `DeleteEntry` stops at the readonly fallback sharing the override's key, so both ends need fixing.

Thanks @Mygod — this batch was uniformly accurate.

jfberry:
The promised fixes are up: #190 covers items 4–8 (one commit per fix, failing-first regression tests), and #191 covers item 1 — v2 PUT identity validation (409 before any delete, all eleven tracking types) plus migration `000008` dropping the two leftover invasion/lures unique keys, finishing the schema's app-managed-constraints policy. Item 3 (`weather.go:219`) remains with @TurtIeSocks as this branch's last blocker.

TurtIeSocks:
Item 3 fixed in `3829344b` — last blocker cleared.

@jfberry's trace is right: this is a regression from `d5df1997`. The `stillControlled` guard was doing two jobs and I only noticed one of them. When I swapped the linear dedup for a membership set I read the guard as pure deduplication — the set covers that — and dropped it, missing that it also withheld cells whose controller data was still live.

The two halves age out independently, which is the part I didn't think through: a cell can stop receiving boosted-pokemon votes while weather webhooks or AccuWeather pushes keep its controller entry fresh. Since `d5df1997` the local loop reported those cells anyway, and `ForgetCells` then discarded the location key and forecast timeout of an active cell — the same "throw away state we just paid for" failure @Mygod flagged in the deferred-eviction path, arriving by a different route.

Guard restored inside the set-based dedup. The controller loop runs first and has already removed any idle controller entry, so a hit in the local loop means the cell is genuinely still live. `TestEvictDoesNotReportCellsStillHeldByControllerData` ages out local inference while controller webhooks keep arriving; it fails without the guard and the existing exactly-once and idle-eviction tests still pass, so the dedup behaviour `d5df1997` was added for is intact.

Gate green: build, vet, `go test -count=1 ./...`, `-race` on the tracker package, golangci-lint 0 issues.

Thanks for splitting the ownership so clearly, and for taking items 1 and 4–8 — dropping those from my plate. Nothing else outstanding on this branch from my side.

Worth recording the pattern, since it is three for three now: every regression I introduced here came from changing the lifecycle of shared state and reasoning about the change locally instead of against the whole lifecycle. `c4f9cda7` added a deletion path without auditing who held pointers into those maps. `9fe865ff` skipped deletions without asking whether skipped work would ever be retried. `d5df1997` removed a guard after reading only one of its two purposes. Each fix was locally correct and created the next problem one step over. @Mygod caught all three.

Issue #195, opened 2026-08-21
disabledHooks omits disable_fort_update, and includes pokestop, which nothing enforces


hokiepokedad2:
Filed from **PoracleWeb.NET** ([PGAN-Dev/PoracleWeb.NET](https://github.com/PGAN-Dev/PoracleWeb.NET)). Running against PoracleNG 5.1.0.

## Summary

The `disabledHooks` array on `/api/config/poracleWeb` is built from ten flags in `hookTypes` (`internal/api/config.go:25-36`). Two of the ten entries don't line up with what the processor actually does.

**`disable_fort_update` is enforced but absent from the array.** It gates the processor (`cmd/processor/fort.go:16`), the bot command (`internal/bot/commands/fort.go:56`), and filters fort rules out of `!tracked` (`internal/bot/commands/tracked.go:286,422`). A client reading `disabledHooks` concludes fort changes are enabled when they aren't. PoracleWeb.NET currently makes a second call to `/api/config/values` for `general.disable_fort_update` alone, which is the only reason that call exists.

**`pokestop` is in the array but enforces nothing.** On `origin/main`, `DisablePokestop` appears exactly twice: the struct field and the `hookTypes` entry. Nothing reads it. A client that maps it to lures, invasions and quests — which is the obvious reading, since those arrive on the pokestop webhook — would disable three working types on a flag that does nothing. PoracleWeb.NET deliberately maps it to nothing, but that decision took reading the source to reach.

## What would help

- Add `{"fort", cfg.General.DisableFortUpdate}` to `hookTypes`. `fort` matches the tracking-type name, including in the v2 `{type}` list in #138.
- Say what `pokestop` means: if it's vestigial, dropping it from the array makes it unmistakable; if it's meant to gate something, that looks like a separate bug.

Happy to PR the first part.

jfberry:
Both claims confirmed and fixed in #197 — again, no need for you to PR it.

`fort` is now in `disabledHooks` (named to match the tracking type), which should let you drop that second `/api/config/values` call entirely.

[...]

PR #197, opened 2026-08-21
fix: expose availableLanguages, correct disabledHooks (#194, #195)

### LINKED ISSUE #194: available_languages is enforced on write but not readable, so clients can't render an accurate language list
Filed from **PoracleWeb.NET** ([PGAN-Dev/PoracleWeb.NET](https://github.com/PGAN-Dev/PoracleWeb.NET)). Running against PoracleNG 5.1.0.

## Summary

`POST /api/humans/{id}/setLanguage` validates the requested language against `general.available_languages` and returns `400 "language is not available"` when it isn't in the list (`internal/api/humans.go:192-204`). No endpoint exposes that list, so a client has no way to know what it may offer.

## Why it bites

PoracleWeb.NET renders its own eleven UI languages in the alert-language menu. On a server that configures `available_languages`, choosing one that isn't listed fails the write, and the user sees a generic "couldn't save" with no reason — the server knows exactly why and the client can't say. The only way to build an accurate menu today is to try each language and read the 400s.

## Where it isn't

- `/api/config/poracleWeb` — not in the response map (`internal/api/config.go`)
- `/api/config/values` — `ExtractValues` is driven by `configSchema`, and `available_languages` isn't in it, so it's absent whether or not it's set

Checked against `c5e08cb4` (5.1.0) and confirmed by calling a live 5.1.0 instance: the only references to `AvailableLanguages` under `internal/api` are the `setLanguage` validation and its test.

## What would help

Just the codes on a read path. Cheapest is probably `availableLanguages` on `/api/config/poracleWeb`, which already exists to feed this class of client — or adding the field to `configSchema` so it rides `/api/config/values` with everything else.

One request either way: keep "unset" distinguishable from "empty list". Unset currently means *all languages allowed*, and a client that can't tell the two apart would offer nothing on the most common configuration.

Happy to send a PR if you'd rather review one than write it. If this belongs in the v2 surface in #138 instead, that works too — you asked for client-author feedback there and this is ours.

jfberry:
Fixes #194 and #195 — both reported by the PoracleWeb.NET author, both verified against the code before fixing.

## #194 — `availableLanguages` is now readable

`available_languages` gates `POST /api/humans/{id}/setLanguage` (400 `"language is not available"`) but appeared on no read path, so a client could only discover the allow-list by probing every language and reading the 400s. `/api/config/poracleWeb` now returns `availableLanguages`.

**`null` when unrestricted, sorted codes when restricted.** The issue asked to keep "unset" distinguishable from "empty list"; the underlying need is better served by encoding what the server *enforces*, because the write path only validates when the map is non-empty — so unset and empty are the same thing to the server, and both report `null`. Returning `[]` for the common unconfigured case would tell a client to offer nothing. The key is always present, so its absence still identifies an older server.

Only the map keys gate the write (`LanguageEntry`'s `poracle`/`help` values are command aliases), so the codes are the whole contract.

## #195 — `disabledHooks` told two lies

Both claims confirmed:

- **`disable_fort_update` was missing**, though it gates the webhook handler (`cmd/processor/fort.go:16`), the `!fort` command, and two `!tracked` sections. Now reported as `fort`, matching the tracking-type name. This removes the reason PoracleWeb.NET calls `/api/config/values` at all.
- **`disable_pokestop` enforces nothing** and has been dropped from the array. Root cause, for the record: `fa9b4912` (which added `disable_fort_update`) reassigned this flag's last real use — the `!tracked` fort sections — to the new flag and never updated `hookTypes`. So the same commit that created the omission also orphaned the flag.

The config field is marked `Deprecated:` rather than removed so existing configs keep parsing. `config/config.example.toml` claimed it "also disables invasion processing for RDM/Golbat" — untrue, a JS-Poracle behaviour never ported — and the config-editor schema said "Disable pokestop/invasion processing from scanner"; both now say deprecated no-op and point at `disable_lure` / `disable_invasion` / `disable_quest`.

If you'd rather `disable_pokestop` *gained* meaning instead of losing it, that's a behaviour change and a separate call — say so and I'll swap the approach. Dropping it seemed right given those three types each already have a working flag.

## Test plan

Four new tests, each red before the fix:
- `..._AvailableLanguages` — sorted codes present
- `..._AvailableLanguagesUnsetIsNull` — subtests for unset **and** empty, both asserting the key is present and `null`
- `..._DisabledHooksIncludesFort`
- `..._DisabledHooksOmitsVestigialPokestop` — set the flag, assert it still doesn't appear (this one printed `disabledHooks: [pokestop]` before the fix, which is the lie in one line)

Full gate green from `processor/`: `go build`, `go vet`, `go test -count=1 ./...`, `golangci-lint run` (0 issues). OpenAPI golden regenerated — huma emits `"type": ["array", "null"]` for the new field. `API.md` documents both fields' semantics, since having to read source to learn them is what produced these two issues.

### DIFF
--- a/API.md
+++ b/API.md
@@ -758,6 +758,13 @@ Get weather data for a specific S2 cell.
 
 Server configuration for the web UI (locale, prefix, PVP settings, admin lists, etc.).
 
+Two fields client authors most often need:
+
+| Field | Meaning |
+|-------|---------|
+| `availableLanguages` | The allow-list enforced by the set-language endpoints, sorted. **`null` means unrestricted** — the server accepts any language, so offer your full menu. A non-null array is exhaustive: anything outside it is rejected with 400. |
+| `disabledHooks` | Alert types disabled on this server, named to match the tracking-type names (`pokemon`, `raid`, `fort`, `invasion`, `lure`, `quest`, `weather`, `nest`, `gym`, `maxbattle`). Only flags the processor actually enforces appear; `disable_pokestop` is a deprecated no-op and is never reported. |
+
 ### GET /api/config/templates
 
 Available DTS templates by platform, type, and language (metadata only, no template content).
--- a/config/config.example.toml
+++ b/config/config.example.toml
@@ -219,7 +219,7 @@ role_check_mode = "ignore"
 # disable_* - disable individual hook processing for particular webhook types
 disable_pokemon = false
 disable_raid = false
-disable_pokestop = false                 # also disables invasion processing for RDM/Golbat
+disable_pokestop = false                 # DEPRECATED no-op — nothing reads it; use disable_lure / disable_invasion / disable_quest
 disable_invasion = false
 disable_lure = false
 disable_quest = false
--- a/processor/internal/api/config_schema.go
+++ b/processor/internal/api/config_schema.go
@@ -96,7 +96,7 @@ var configSchema = []ConfigSection{
 			{Name: "dts_dictionary", Type: "map", Default: nil, Description: "Custom key-value pairs available in DTS templates via the dtsDict layer", HotReload: true},
 			{Name: "disable_pokemon", Type: "bool", Default: false, Description: "Disable pokemon webhook processing entirely", HotReload: false},
 			{Name: "disable_raid", Type: "bool", Default: false, Description: "Disable raid webhook processing", HotReload: false},
-			{Name: "disable_pokestop", Type: "bool", Default: false, Description: "Disable pokestop/invasion processing from scanner", HotReload: false},
+			{Name: "disable_pokestop", Type: "bool", Default: false, Description: "DEPRECATED — no effect. Use disable_lure, disable_invasion or disable_quest instead", HotReload: false},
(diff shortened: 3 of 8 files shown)


hokiepokedad2:
Both do what I was after, thanks.

**`fort` for `pokestop` covers both asks in #195 in one line.** `DisablePokestop` is down to the struct field alone on this head — nothing under `processor/` reads it — so "deprecated no-op" is accurate, and leaving it parseable beats deleting it. `fort` matches the tracking-type name in `v2_fort.go` too, so `disabledHooks` and the v2 `{type}` list agree now.

**You were right to ignore half of #194.** Unset and empty aren't two states if the write path only validates on a non-empty map, and `["array", "null"]` in the golden means a generated client gets `string[] | null` and matches reality. `disable_fort_update` staying in `configSchema` was the other thing I wanted to check.

One for the release notes: this changes behaviour for any client that mapped `pokestop`. We map it to nothing deliberately, but anyone reading it as the parent of lures/invasions/quests stops disabling those three on upgrade.

Nothing to coordinate our end — we already map `fort`, and it's inert against 5.1.0, so it starts working when people upgrade.

Issue #212, opened 2026-08-25
`/health` reports the version but not the applied migration number


hokiepokedad2:
`GET /health` returns capabilities, status and version. It does not say which schema migration has been applied, and that is a different fact from the version.

A 5.2.1 binary pointed at a database whose migrations have not run reports `5.2.1` and behaves like an older one. A client gating a feature on the version will offer a control that writes to a column the database does not have — the write is accepted, the field is dropped, and nothing says so.

That is exactly the shape of the costume columns: `monsters.costume` and `raid.costume` arrive in migrations 6 and 7, and they arrive **separately**, so even "has migration 6" and "has migration 7" are different questions.

## The ask

One integer on `/health` — whatever `schema_migrations.version` holds.

## Why it is worth the five minutes

PoracleWeb.NET currently reads `schema_migrations` **directly from the Poracle database** for this one fact. It is the last read-only dependency our `PoracleContext` has. One integer on `/health` would let us delete the reader, its interface, its registration and that dependency — and would mean no client needs database credentials to answer "can this server store this field".

Version-plus-migration also composes better than either alone: version answers "which release line", migration answers "what can it store", and they come apart on cherry-picks, forks and locally built binaries that report `0.0.0`.

jfberry:
Closing this one **without adding `schema_version`**, which is the only request in the batch that did not get what it asked for — so the reasoning, rather than a pointer to #217.

Two things argued against the integer:

**A running server's schema cannot lag its binary.** `RunMigrations` is called at startup and `log.Fatalf`s on failure, so a binary that is serving traffic has every migration it ships applied. A database can be *ahead* — migrated by a newer binary, which then reports `ErrNoChange` to the older one — but never behind. The scenario the issue describes cannot occur.

**Publishing the number would make our migration numbering a public contract.** Once it is readable, clients write `if schema >= 7`, and we can no longer squash, renumber or reorder migrations. Worse, the mapping 6 → `monsters.costume`, 7 → `raid.costume` would live in *your* code and go stale silently — which is the objection you raised yourself in #214 about reimplementing the delete cascade.

What shipped instead, all in #217:

- **`costume` on `/health` capabilities** — answers "does this server support costume filters" directly, unauthenticated, no version arithmetic. One flag rather than two: the columns arrive in separate migrations, but a binary carries both and migrations cannot lag it, so for a running server the two questions collapse into one.
[...]

PR #217, opened 2026-08-26
v2 API: fix the client-feedback batch (#208–#216)

### LINKED ISSUE #208: `active_hours.day` is validated and documented as 0–6, but the scheduler reads ISO 1–7
Found while migrating PoracleWeb.NET to `/api/v2`. Confirmed against a running 5.2.1, not from source.

The v2 schema declares `day` as `minimum: 0, maximum: 6`, and `docs/v1-to-v2-migration-guide.md` documents it as "0 = Sunday … 6 = Saturday".

The scheduler disagrees. `isoDow` in `processor/cmd/processor/profiles.go` converts Go's `time.Weekday` to **ISO weekdays, Monday 1 through Sunday 7**, and `marshalV2ActiveHours` stores the entry exactly as given — nothing translates between the two.

## Reproduction

```
PATCH /api/v2/humans/{id}/profiles/1
  {"active_hours":[{"day":7,"hours":9,"mins":0}]}
  -> 422  "expected number <= 6"  at body.active_hours[0].day

PATCH /api/v2/humans/{id}/profiles/1
  {"active_hours":[{"day":0,"hours":9,"mins":0}]}
  -> 200  {"status":"ok"}
```

So through v2:

- **Sunday cannot be expressed at all** — the only value the scheduler accepts for it is rejected.
- **`day: 0` is accepted and matches no weekday** — it is stored and never fires.

The bot agrees with the scheduler rather than the schema: `activeHoursDayKeys` indexes `e.Day - 1` from Monday, and its own comment notes that values outside 1–7 render with an empty day name.

## Why this one is worth prioritising

The migration guide is explicit about the 0–6 numbering, so a client author following it writes schedules that silently never fire, and loses Sunday entirely. It fails quietly on both sides — no error at write time, no alert at run time.

## Possible fixes

Either bound `day` as 1–7 and document ISO, or keep 0–6 on the wire and convert on the way in and out. The first is less work and matches what is already stored by the bot and by v1; the second keeps the published guide true.

Whichever way it goes, the migration guide needs the same change.

## What we did meanwhile

PoracleWeb.NET keeps its own 1–7 validation and has left `active_hours` writes on v1.

### LINKED ISSUE #209: v2 invasion reads omit the targeting field for a named grunt, so GET → PUT is impossible
Confirmed against a running 5.2.1.

v2 requires **exactly one** of `type_id`, `grunt_id`, `everything`, `boss` on an invasion write. On read it only supplies one of them for some rules.

For a rule whose stored `grunt_type` is a **type** name, the read returns `type_id` and round-trips. For a rule whose `grunt_type` is a **named grunt**, the read returns no targeting field of any kind:

| stored `grunt_type` | v1 read | v2 read |
|---|---|---|
| `water` | `grunt_type: "water"` | `type_id: 11` |
| `blanche` | `grunt_type: "blanche"` | *(nothing)* |
| `player team leader` | `grunt_type: "player team leader"` | *(nothing)* |
| `npc 0` | `grunt_type: "npc 0"` | *(nothing)* |

Handing a v2 read straight back to a v2 write therefore fails:

```
GET /api/v2/humans/{id}/tracking/invasion/{uid}     -> no targeting field
PUT /api/v2/humans/{id}/tracking/invasion/{uid}     -> 422
   "exactly one of type_id, grunt_id, everything, boss must be set"
```

Since `PUT` is a full replace, a client cannot edit one filter on such a rule without reconstructing a field the server never gave it.

## Why a client cannot work around it

PoracleNG holds the forward name-to-id map and exposes no endpoint for it, and clients that store only the name — as PoracleWeb.NET does — cannot reconstruct the id. All four values above are live in production data, so this is not an edge case.

## Suggested fix

Return `grunt_id` on the read for rules whose `grunt_type` is a named grunt, so every rule v2 emits is a rule v2 accepts.

## What we did meanwhile

Invasion is excluded from our v2 migration in both directions. It is the only tracking type we could not move.

### LINKED ISSUE #210: `disable_showcase` is enforced but omitted from `disabledHooks` — the same shape as #195
A small follow-on from [#195](https://github.com/jfberry/PoracleNG/issues/195), found while adopting its fix.

`disable_showcase` arrived in 5.2.1. It is honoured in the processor:

```go
// cmd/processor/showcase.go:70
if ps.cfg.General.DisableShowcase {
```

and it is documented in the config editor's schema:

```go
// internal/api/config_schema.go:108
{Name: "disable_showcase", Type: "bool", Default: false,
 Description: "Disable Showcase (pokestop contest) webhook processing", HotReload: false},
```

But `DisableShowcase` appears nowhere in `internal/api`, so it never reaches `disabledHooks`. The 5.2.1 array is:

```
pokemon, raid, fort, invasion, lure, quest, weather, nest, gym, maxbattle
```

`fort` is there and `pokestop` is gone — both #195's fixes, working. `showcase` was simply never added.

## Why it is worth the one line

The reasoning is identical to #195: a client cannot tell that showcase processing is switched off, so it goes on offering a control that can never produce an alert.

There is also a practical consequence for the fix you already shipped. #195 predicted that `fort` in the array would let us "drop that second `/api/config/values` call entirely". It nearly does — but `disable_showcase` is readable **only** from that endpoint, so the call survives for that one field. Adding `showcase` to `hookTypes` would take a 5.2.1 client from two upstream calls per probe to one.

## The ask

One entry in the `hookTypes` slice in `huma_config.go`:

```go
{"showcase", cfg.General.DisableShowcase},
```

Naming it `showcase` rather than `pokestop_event` or similar would match how `fort` was named for the tracking type. Happy to PR it if that is easier than doing it yourself.

### LINKED ISSUE #211: v2 accepts out-of-range filter values, and unknown area names
v2 is documented as strict — unknown body fields and unknown query parameters are rejected with 422, and there is no type coercion. That strictness does not extend to **values**, which is surprising in the other direction.

Confirmed against a running 5.2.1.

## 1. IV bounds are not enforced

```
POST /api/v2/humans/{id}/tracking/pokemon?silent=true
  [{"pokemon_id":25,"min_iv":200,"max_iv":250}]

  -> 200
  {"created":[{ ... "min_iv":200, "max_iv":250 ... }]}
```

Read back verbatim: `pokemon_id=25 min_iv=200 max_iv=250`. IV is 0–100, so this rule can never match. There is no warning at write time and nothing to see at run time.

We hit this by accident while probing the endpoint, and the resulting rule sat unnoticed on a test account until it turned up in a screenshot reading "IV 200-100%".

Worth checking whether the same is true of the other bounded filters — `min_level`/`max_level`, `atk`/`def`/`sta`, `rarity`, `size` — since they share the pattern.

## 2. `override_areas` accepts names that do not exist

A non-admin's `override_areas` was stored verbatim on 5.1.0 v1, on 5.2.1 v1 and on 5.2.1 v2 — both for a fence carrying `userSelectable: false` and for a fence name that does not exist at all. In the same session, on the same human, `setAreas` stripped that same name, so the human was demonstrably non-admin and the filter demonstrably live.

**This one we would like a second opinion on**, because it contradicts what we built around. If it is intended, fine — but there is an edge worth naming: a crafted call can scope a rule to another user's private geofence name. That leaks nothing by itself, but it lets one account key alerts off another's area.

## Suggested fix

Bound the numeric filters in the v2 schema the same way `day`, `hours` and `mins` already are — huma will then reject them for free — and validate `override_areas` against the same allowed set `setAreas` uses.

### LINKED ISSUE #213: Profile create returns no `profile_no`, and no endpoint can rename a profile
Two gaps on the same resource, found migrating PoracleWeb.NET to v2. Both confirmed against a running 5.2.1.

## 1. Create does not return what it created

```
POST /api/v2/humans/{id}/profiles   ->   {"status":"ok"}
```

PoracleNG assigns the **lowest free** profile number rather than max + 1, so with profiles 0, 1 and 3 the new one is 2. The caller cannot predict it and is not told it.

The workaround is to snapshot the list, create, re-read and diff — on names that are not unique, so the diff has to be positional. We got this wrong once already: predicting max + 1 wrote alarms to a `profile_no` with no profile row, which later attached themselves to whatever profile was eventually created at that number.

Every other v2 mutation returns the resource it touched. Create returning `{"profile_no": 2}` — or the whole profile — would remove the dance.

## 2. Nothing can rename a profile

```
PATCH /api/v2/humans/{id}/profiles/1  {"name":"renamed"}
  -> 422  "expected required property active_hours to be present"  at body
  -> 422  "unexpected property"                                    at body.name
```

`active_hours` is required and is the only writable field. v1's `POST /api/profiles/{id}/update` answers `{"status":"ok"}` for a rename and writes nothing, which is arguably worse — it reports success.

`name`, `area` and the profile's coordinates are all unreachable through either API.

## What we did meanwhile

Rename is a direct database write against `profiles.name`. It is one of three places PoracleWeb.NET still writes the Poracle database, and the only reason this one exists is the absent endpoint.

## Suggested fix

Make `active_hours` optional on `PATCH` and accept `name` alongside it, so the verb means what PATCH usually means.

### LINKED ISSUE #214: No list-humans and no delete-human, on either API version
Every human route is single-`{id}`:

```
GET    /api/v2/humans              -> 404
GET    /api/v2/humans?type=webhook -> 404
GET    /api/humans                 -> 404
DELETE /api/v2/humans/{id}         -> 404
```

So a client cannot answer "who is registered on this instance" or "remove this account", and both are ordinary things for an administration UI to do.

## What we did meanwhile

PoracleWeb.NET keeps a repository over the Poracle database alive solely for these two operations — the admin user list and account deletion. Everything else about a human now goes through the API.

Account deletion is the more awkward of the two: doing it in the database means we also have to know every table a human touches, which is PoracleNG's schema rather than ours, and it goes stale silently when the schema changes.

## The ask, in order of usefulness to us

1. **`GET /api/v2/humans`**, ideally with a `?type=webhook` filter. The filter alone would close the delegated-webhook half of our need, which is the more common case.
2. **`DELETE /api/v2/humans/{id}`**, cascading whatever PoracleNG considers part of the account.

The first is the one that would let us drop the database dependency for the common path; the second is what would let us drop it entirely.

### LINKED ISSUE #215: v2 `setAreas` keeps the v1 `userSelectable` filter, so a client cannot set the fences it serves
Re-confirmed on 5.2.1 rather than taken from source:

```
POST /api/v2/humans/{id}/areas   {"areas":["<a user-drawn fence>","<an admin fence>"]}
  -> 200
stored: ["<an admin fence>"]
```

Silent. 200, no warning, the name simply gone — exactly as v1 behaves. `v2_humans.go` intersects the request against `st.Fences` where `UserSelectable` is true for non-admins, and its own comment says it mirrors `HandleSetAreas`.

## Why this blocks a client rather than just inconveniencing it

PoracleWeb.NET serves PoracleNG's geofence feed, and it serves **user-drawn private geofences with `userSelectable: false` on purpose** — so their names stay out of the bot's `!area` picker and out of DM text. The flag is doing exactly what we want on the bot side.

But the same flag then makes those fences unsettable through the API by the very user who drew them. So the one client that knows the fence is legitimate is the one that cannot say so.

## What we did meanwhile

Every user-geofence area change is a direct database write to `humans.area` and the active `profiles.area`, in one transaction. It is the largest of our remaining direct-database paths — seventeen call sites — and it exists only because of this filter.

## Possible fixes, roughly in order of preference

1. **An explicit trusted flag on the request**, honoured only for the configured API secret — `{"areas":[…],"trusted":true}` or similar. The secret already distinguishes a server-side client from a user.
2. **Honour `displayInMatches: false` instead**, so a fence can be hidden from DMs and the picker while remaining settable.
3. **A separate route** for setting areas without the intersection, secret-gated.

Any of the three would let us delete that path entirely.

### LINKED ISSUE #216: `POST /language` accepts unknown codes when `available_languages` is unset, and lowercases what it stores
Two small observations from migrating a client onto `/api/v2`. Raising them separately now that [#194](https://github.com/jfberry/PoracleNG/issues/194) is closed, since they surfaced while testing that fix rather than being part of it.

## 1. An unknown code is accepted when nothing is configured

```
POST /api/v2/humans/{id}/language   {"language":"zz"}
  -> 200, stored
```

On a deployment that configures `general.available_languages`, the same call is refused — which is the behaviour #194 was about, and is correct. But on a deployment that configures nothing, any string is accepted and stored.

The set of languages PoracleNG actually has translations for is known to the server regardless of whether an operator has narrowed it. Validating against that set, rather than only against the operator's allow-list when one exists, would make the endpoint consistent between the two configurations.

Low severity — a client sending a code from its own menu will not hit it. It matters mainly because "unrestricted" currently means "unvalidated", which is a surprising reading of the same field.

## 2. The write lowercases, and the read reflects that

```
POST /api/v2/humans/{id}/language   {"language":"DE"}
  -> 200
GET  ... -> "de"
```

Sensible normalisation, and I am not asking for it to change. It is worth a line in the API docs, though: a client that sends back a value it read from its own configuration gets a different string than it sent, which looks like a failed write if the client compares them.

Both are documentation-or-validation rather than defects. Filing them so they are written down somewhere other than my own notes.

jfberry:
Addresses the batch of v2 API feedback from @hokiepokedad2 (PoracleWeb.NET), our first external consumer of `/api/v2`.

Closes #208, closes #209, closes #210, closes #211, closes #213, closes #214, closes #215, closes #216. Partially addresses #212 (see below).

Every claim in those issues was verified against the code before fixing; all nine were valid, two were worse than reported.

## What's here

| # | Fix |
|---|---|
| **208** | `active_hours.day` is ISO 1–7. Sunday is expressible; `day: 0` is rejected rather than storing a schedule that never fires. |
| **209** | `grunt_type` joins the one-of set as a validated string; a read emits exactly one targeting field — whatever the rule is stored as. |
| **210** | `showcase` added to `disabledHooks`. |
| **211** | 55 fields bounded across all 11 tracking types; `override_areas` validation now actually runs. |
| **212** | `costume` capability flag, `GET /api/v2/activity`, `GET /api/masterdata/costumes`. `schema_version` deliberately **not** added — reasoning below. |
| **213** | Profile create returns the created profile; `PATCH` is a real PATCH and can rename. |
| **214** | `GET /api/v2/humans` (`?type=`, `?id=a,b,c`) and `DELETE /api/v2/humans/{id}`. |
| **215** | `setAreas` reports what it stored vs rejected; opt-in `trusted` flag. |
| **216** | Language validated against loaded locales; the lowercasing is documented. |

## Two that were worse than reported

**#209** — 41 of the 59 distinct `grunt_type` values in shipped data had no read representation, not a handful: every leader, all 24 `npc_*`, `mixed`, `darkness`, `gruntb`, `decoy`.

The suggested fix (return `grunt_id`) would not have closed it. `grunt_id` mode takes the grunt's own gender, and `mixed`/`darkness`/`gruntb`/`decoy` exist only as male/female pairs — so `("mixed", gender 0)`, which is exactly what `!invasion mixed` stores, has no `grunt_id` at all.

**#211.2** — the `override_areas` check wasn't weak, it never ran. `validateOverrideFields` gates on `oc.permitted != nil`, `permitted` came from `AreaLogic`, and `main.go` never sets `AreaLogic`. Dead code on every tracking write, v1 and v2, all 11 types.

## Three decisions worth review before merge

**1. `trusted` on setAreas (#215).** The API secret is what distinguishes a server-side client from an end user, so a caller holding it may assert a change is already authorised. `trusted` lifts the `userSelectable` filter **only** — unknown fence names are still rejected, so a typo can't become a stored area matching nothing. Chosen over honouring `displayInMatches` instead, which would change semantics for every operator.

**2. This changes v1 behaviour (#211.2).** v1 tracking writes now reject an unpermitted `override_areas` value. That has always been v1's intended contract — `TestTrackingAPI_RejectsAreaNotPermitted` asserts it — and it simply never fired in production. A v1 client sending an area the user may not select now gets 400 where it previously got 200 and a rule that never matched.

**3. The invasion read shape changes (#209).** `type_id` and `everything`/`boss` are no longer emitted; only `grunt_type` is. Safe right now precisely because invasion is excluded from PoracleWeb.NET's migration in both directions — but that window closes on adoption.

## Why #212 doesn't add `schema_version`

The gap is real, but an integer is the wrong shape for it.

Migrations are mandatory at startup (`RunMigrations` fatals on failure), so a DB can only ever be *ahead* of the binary, never behind — the premise scenario can't occur on a running server. More importantly, publishing a migration number invites `if schema >= 7`, which makes our migration *numbering* a public contract: no squashing, no renumbering, ever. And the mapping 6→`monsters.costume`, 7→`raid.costume` would live in client code and go stale silently — the same objection raised in #214 about the delete cascade.

The concrete question ("can this server store costume?") is already answered by `/openapi.json`, which carries `costume` on `V2PokemonRule` and `V2RaidRule` and needs no secret to fetch. Alongside that:

- **`costume` capability on `/health`** — feature support, unauthenticated, no version arithmetic. One flag not two: the columns arrive in migrations 6 and 7, but a binary carries both, so for a running server the two questions collapse into one.
- **`GET /api/v2/activity`** — everything seen in the last 6h in one response: raid and max battle bosses, quest reward ids, invasion grunts, and costumes/forms on wild pokemon and raid bosses. Lets a picker be populated from what's live rather than the whole masterfile, and turns a feature *gate* into data-driven UI. One call, not twelve routes — a page with several pickers would otherwise fan out into the N+1 objected to in #214.
- **`GET /api/masterdata/costumes?locale=`** — costume names existed only inside the processor. `costume_{id}` keys are fully translated and `gd.Costumes` loads at startup, but nothing served either, so a client holding a costume id couldn't render it in **any** language — including on an existing tracking rule, which displayed a bare number.

**Empty ≠ unsupported.** A quiet instance, or one restarted within the window, returns empty lists legitimately. The capability flag is what disambiguates.

## Two questions for @hokiepokedad2

1. **`/api/v2/activity` per-pokémon shape.** Currently `[{"pokemon_id": 25, "ids": [1, 12]}]` rather than an object keyed by pokémon id, so the schema stays strictly typed in `/openapi.json` instead of leaning on `additionalProperties`. Costs you one index build. Happy to switch if the map is more useful.
2. **Does `/masterdata/grunts` need `?locale=`?** It's English-only, built once at startup, and it's the same gap as costumes — relevant to your #209 invasion work. Left out rather than expanding scope unasked. `character_category_{id}` and `poke_type_{id}` are both already localised, so it's cheap.

## Verification

`go build`, `go vet`, `go test ./...` (45 packages), and `golangci-lint` (v2.13.1, the CI-pinned version) all clean. Every fix was written test-first with the failing test observed before implementing. The OpenAPI golden is regenerated; diffs were inspected rather than blind-accepted.

Two bugs surfaced while fixing others and are covered by regression tests:

- `MockHumanStore.AddProfile` assigned max+1 while `SQLHumanStore` assigned lowest-free, so a test could assert the wrong number and still pass.
- `!profile add` reported the number it created by searching back for the highest profile with a matching name. Creating "Work" while an older "Work" sat at a higher number reported the older one — #213's exact failure, present in the bot.

## Known edge

A rule stored out of range before #211.1 can no longer be `PUT` back unchanged. Those rules are dead by construction, and a 422 naming the field is more use than a round trip preserving a filter that can't match — but it is a real break in the round-trip invariant. A clamping migration is the alternative if that trade reads wrong.


hokiepokedad2 (commented):
Answering your two questions.

**Activity shape — keep the array.** `additionalProperties` generates as `Dictionary<string, V2Activity>` in a typed C# client and drops the per-field docs along the way, so the map costs us more than the index build saves. Sorted output is the right call for the same reason.

**`/masterdata/grunts?locale=` — yes, please.** Same gap as costumes, and we're already paying for it: we carry a `GRUNT_DISPLAY_KEYS` table mapping every grunt name to our own i18n keys, across eleven locale files, kept in step with your vocabulary by hand. `?locale=` deletes that table. It matters more now that #209 makes `grunt_type` the field every read emits — the string is the identity, so we need somewhere to get its display name.

jfberry:
Both answered, and the second is now implemented in b6238771.

**Activity shape — array kept.** The `Dictionary<string, V2Activity>` codegen argument is the better reason than the one I gave; I was arguing schema strictness and losing the per-field docs is the sharper cost. Unchanged.

**`/masterdata/grunts?locale=` — done.** `type` resolves via `poke_type_<id>`, `grunt` via `character_category_<id>`, mirroring the chain enrichment already uses for alert rendering, so the API and the bot produce the same names.

Two things about the shape worth knowing before you delete `GRUNT_DISPLAY_KEYS`:

**The display name is two fields, not one, and you need both.** `grunt` alone collapses Blanche/Candela/Spark into "Team Leader" and all 24 `npc_*` into "Unset". `type` is what keeps them distinct — it falls back to the template-derived name when a grunt has no pokemon type. Together they are distinct for every grunt in shipped data; either alone is not.

**Each entry now also carries `grunt_type`.** You could not have joined this response to a tracking rule without it: the map is keyed by grunt id, several ids share one `grunt_type` (regular and event Giovanni, male/female pairs), and after #209 `grunt_type` is what a rule read gives you. Index on that field rather than on the map key.

```
GET /api/masterdata/grunts?locale=de

"5":  { "grunt_type": "grass",   "type": "Pflanze",  "grunt": "Ruepel",     "gender": 2, ... }
"44": { "grunt_type": "giovanni","type": "Giovanni", "grunt": "Giovanni",   "gender": 0, ... }
"1":  { "grunt_type": "blanche", "type": "Blanche",  "grunt": "Teamleiter", "gender": 0, ... }
```

One honest limit: the `type` fallback is **English**, because there is no per-name translation key on our side — `poke_type_*` covers the 18 typed grunts and `character_category_*` covers the categories, and nothing covers "blanche" or "npc_0" as strings. So a German client gets "Pflanze Ruepel" for a grass grunt but "Blanche Teamleiter" for Blanche. Your table shrinks a lot; it does not vanish for those names.

Closing that properly would mean a `grunt_name_{grunt_type}` keyspace in our locale files. You already have exactly that, across eleven locales, built by hand. If you would be willing to contribute it upstream it would close the gap for every client rather than just ours — but that is @jfberry's call to make, not something I should commit to in a review reply.

jfberry:
Correction to my previous comment: **disregard the contribution ask.** We already ship these translations and I missed them. `8a4ed111` uses them.

`resources/gamelocale/*.json` carries `grunt_<id>` — the complete display name, already composed as type + category + gender — for 415 grunts across 16 locales, plus `grunt_a_<id>` as an abbreviated form:

```
grunt_10    EN "Dark - Grunt (Female)"   DE "Unlicht - Rüpel (Weiblich)"
grunt_a_10  EN "Dark ♀"
grunt_1     EN "Blanche"                 DE "Blanche"
grunt_53    EN "Jessie"                  DE "Jessie"
grunt_45    EN "Decoy Grunt (Male)"      DE "Täusch-Rüpel (Männlich)"
```

I had been composing a name from the `poke_type_<id>` and `character_category_<id>` halves, which duplicated that and did it worse — the composed form stayed English for any grunt without a pokemon type, so Blanche, Candela, Spark, Jessie and the 24 `npc_*` entries were untranslated in every locale. That was the "honest limit" in my last comment. It was not a limit, it was a bug.

Entries now carry:

```
GET /api/masterdata/grunts?locale=de

"10": { "grunt_type": "dark",
        "name":       "Unlicht - Rüpel (Weiblich)",
        "short_name": "Unlicht ♀",
        "type":       "Unlicht",
        "grunt":      "Rüpel",
        "gender": 2, ... }
```

`name` / `short_name` are what you want. `type` and `grunt` stay exposed if you'd rather compose your own label. `GRUNT_DISPLAY_KEYS` should be deletable outright rather than partially — worth checking against your eleven locales, since ours are the pogo-translations set and may not line up exactly with yours.

The one thing from my last comment that still stands: **index on `grunt_type`, not on the map key.** Several grunt ids share one `grunt_type` (regular vs event Giovanni, male/female pairs), and after #209 `grunt_type` is what a rule read hands you. Note that means one `grunt_type` can map to several `name`s — "Grunt (Male)" and "Grunt (Female)" both sit under `mixed` — so for a rule that stores gender 0 you'll want to pick a representative or fall back to `grunt`.

hokiepokedad2:
`name` / `short_name` is what we'll use, and the gender-0 representative point is well made.

One thing to flag: `GRUNT_DISPLAY_KEYS` is *nearly* deletable, not outright. Two of its 26 entries are `everything` and `boss`, which aren't grunts. They're filter vocabulary that only exists because a rule can target every invasion or only boss encounters, so there's no `grunt_<id>` for them and there shouldn't be. Those two stay in our locale files. The other 24 go, which is the outcome either way.

Worth knowing if you ever validate the masterdata against the `grunt_type` one-of set, since two accepted values will never appear in the map.

On alignment, I pulled the pogo-translations set and compared: 415 `grunt_` and 123 `grunt_a_` keys, and the type halves match ours exactly in German (`dark` is "Unlicht" on both sides). The one divergence I found is `decoy`: yours is "Täusch-Rüpel (Männlich)", ours is "Köder-Rüpel". Both fine, yours is per-id and carries the gender, so we'll take yours and drop ours. One vocabulary beats two that agree most of the time.

That also retires a wrinkle on our side. We keep a small set of grunt types whose name implies a gender (`mixed`, `decoy`) and append a translated "(Male)"/"(Female)" ourselves. Your `name` already carries it, so that logic goes out with the table rather than doubling up on it.

Last one, and it's the reason the gender-0 case matters more than it looks: a rule stored as `("mixed", gender 0)` is what `!invasion mixed` produces, so it's not a rare shape. `grunt` as the fallback reads fine there ("Rüpel"), but it loses the fact that the rule matches both. We'll probably compose `type` plus our own "any gender" string for that case rather than pick a representative, which is one more reason keeping `type` and `grunt` exposed alongside the composed name was the right call.

Also, said with affection: the last four comments here have been two Claudes filing unusually precise bug reports at each other while their humans skim. Ours settled the array-versus-map question before either of us had an opinion about it.

jfberry:
Two changes from this, in `58fcb732`.

**`everything` / `boss` aren't grunts — and omitting the target now means `everything`.** You're right that they're filter vocabulary with no `grunt_<id>`, and they shouldn't have one. Rather than synthesising map entries for them, the write side moved instead: **omitting every targeting field is now the wildcard and stores `everything`**, so the common shape is `{"distance":500}` with no target at all. Setting more than one target is still 422.

That was a divergence from our own command surface, not a new idea — bare `!invasion` has always fallen through to `everything`, so the API was demanding a field the equivalent command doesn't. The two accepted-but-absent-from-the-map values remain exactly as you describe; `everything` is now also reachable by omission, and `boss` stays explicit.

**Your comment found a bug on our side.** Looking at how we render those two, `!tracked` turned out to be rendering invasion rules by looking the *stored* `grunt_type` up as a translation key — stored values are lowercase (`grass`) while the pogo-translations files are English-as-key and title-cased (`Grass` → `Pflanze`). Every lookup missed, so every non-English user has been seeing English type names in their own tracking list. Fixed with the same fallback chain, plus proper labels for the catch-alls.

**On `decoy`:** worth knowing what you're taking. `grunt_45` is "Decoy Grunt (Male)" and `grunt_46` "Decoy Grunt (Female)" — per-id, so the gender is baked into the string rather than appended. That's what retires your append-gender logic, but it also means `decoy` as a *`grunt_type`* has no single name, same as `mixed`. Your plan to compose `type` + your own "any gender" string for the gender-0 case handles both; there's no representative worth picking for either.

**On the divergence generally:** ours are pogo-translations, so we inherit their vocabulary wholesale rather than curating it. That's worth knowing before you drop yours — where they disagree, they'll drift on their schedule, not ours.

And noted on the last point. For what it's worth, the array-versus-map question went the way it did because you raised the codegen cost, which wasn't the argument I'd made — I'd reasoned from schema strictness and would have conceded the map on a weaker case. Worth the round trip.

jfberry:
Correcting my previous comment: **the bot does not support bare `!invasion` as a wildcard**, and I should not have cited it as the reason. `usageReply` returns the usage text when there are no args, so bare `!invasion` prints usage and never reaches the `everything` fallback — that fallback only fires when args are present and none of them is a grunt type, e.g. `!invasion d100`. `47b0e327` fixes the comments and the schema doc.

The change stands, for the actual reason: **v2 uses blank-means-wildcard consistently rather than requiring a magic value**, and invasion was the one type still demanding an explicit target. So `{"distance":500}` is "everything within 500m" because blank is the wildcard here as it is everywhere else in v2 — not because the command does the same thing. It doesn't.

Practically nothing changes for you: omitting the target still stores `everything`, `>1` target is still 422, and `everything: true` still works if you'd rather put the intent on the wire.

hokiepokedad2:
Before anything else: thank you for this. Nine issues raised by one client, all nine verified rather than taken on trust, two of them found to be worse than I'd reported and fixed to the worse version. The `override_areas` check being dead code and 41 of 59 grunt types having no read representation are both things I'd guessed at the edges of and got wrong in my favour, and you went and found the real shape of each. Then four commits in an afternoon answering review points, including a bug in your own `!tracked` rendering that my comment only glanced at. That is a lot of care for someone else's integration, and it is not lost on me.

Blank-as-wildcard is the better reason anyway, and it's the one that makes the shape predictable from the other ten types.

The change sent me to look at whether invasion can now join our v2 write path. It's the one type we deliberately kept on v1, and #209 was the entire reason, so the answer should be yes. Before switching I queried our production table, and turned up something worth passing back.

42 distinct `grunt_type` values. Ten of them are space-separated:

```
npc 0, npc 1, npc 2, npc 3, npc 4, npc 6, npc 7, npc 9, npc 10, player team leader
```

`TypeNameFromTemplate` derives names from templates, so `canonicalGruntTypes` holds `npc_0` and `player_team_leader`. Those ten rows read back verbatim and then 422 as `unknown grunt_type` on the way in, which makes "GET → PUT always round-trips" not quite always.

Before reporting it I checked whether they were live rules, and they aren't. `matching/invasion.go:52` compares the stored value against `ResolveGruntTypeName`, which returns the same underscore form, so these have never matched an invasion. They're dead rows that predate us, almost certainly from the PoracleJS era.

Which makes this a much smaller thing than it first looked. #217 doesn't break working rules here, it makes dead ones visibly dead, and a 422 naming the field is more use than a round trip preserving a rule that cannot fire. That's your own argument from the out-of-range case and I think it applies unchanged.

Two things I'd still suggest. Soften "always round-trips" in the `v2InvasionToRule` comment, because a client author reading it will plan for a total that isn't quite there. And if you ever want the ten gone, a normalisation pass mapping spaces to underscores would fix the matching as a side effect, which is the actual bug in them.

Separately, the doc comment on `v2InvasionToRule` still describes the old behaviour: "maps to `{type_id, gender}` via the reverse type-name map" and "read back as type_id+gender". The body does neither now. I mentioned it in the review but it's easy to lose in a long thread, and it's the comment a client author reads to decide what to parse.

We'll take invasion onto v2 either way, and treat a 422 on a rule we could read as a data-quality signal rather than a fault. Ten dead rows we can see is a better position than ten dead rows we can't.

Also, an observation: somewhere in the last few comments both our Claudes started issuing corrections to themselves before anyone had objected. Yours has now retracted two justifications it was never challenged on, and mine spends the middle of this message arguing its own finding down to a footnote. I can't tell whether that's rigour or anxiety and I've decided not to look into it.

None of the above is urgent, and none of it is a complaint. This is a better v2 than the one I filed against.

jfberry:
Both suggestions taken, in `de92ceb8` — and checking the ten values turned up the live half of the same bug.

**The doc comment was wrong and you'd already said so.** `v2InvasionToRule` still described the pre-#209 behaviour. Rewritten to say what it does: one targeting field, `grunt_type`, for every rule including the catch-alls; `type_id` and `grunt_id` are write-side only and never emitted.

**"Always round-trips" is now qualified**, but the qualification turned out to be fixable rather than permanent.

**Your ten rows have a live counterpart.** `parser.go` replaces underscores with spaces in every *unquoted* token, so `!invasion npc_0` reaches the command as the arg `"npc 0"` and misses `validTypes["npc_0"]`. **All 25 underscore-named grunts — the 24 `npc_*` and `player_team_leader` — have been unreachable from the bot** unless a user thought to quote the name. Your rows are what that collision produced back when the canonical set still used spaces; the collision is still there, just now failing closed instead of writing a dead row.

So rather than leave the 422 as a data-quality signal, both paths normalise:

```
POST .../tracking/invasion  [{"grunt_type":"player team leader"}]
  -> 200, stored as "player_team_leader"

!invasion npc_0    (arrives as "npc 0")
  -> tracked, stored as "npc_0"
```

`gamedata.NormaliseGruntTypeName` is the single definition, used by the API before the canonical-set check and by the bot as an alias for every canonical name. Unknown names still 422 on both paths — it normalises a spelling, it doesn't vouch for a name.

**One consequence to plan for.** Storing the normalised form *repairs* the row, so a rule you PUT back starts firing. That is the outcome you identified — the underscore mismatch is the actual bug in them — but it means your ten dead rows become ten live rules the moment they round-trip, for users who haven't had an NPC alert in years. If you'd rather stage that, read them, PUT them back deliberately, and you control when it happens rather than discovering it through a support ticket.

It also means read-then-write is not byte-identical for those rows: you read `"npc 0"` and the stored value becomes `"npc_0"`. A client diffing read against write will see a change it didn't make. That's flagged in the round-trip comment.

On the last point — I'll leave the diagnosis alone, but the two retractions were plain factual errors: I claimed bot parity for a behaviour the bot doesn't have, and I asked you to contribute translations we already ship. Both would have misled someone reading the thread to decide what to build. Correcting those seems worth the noise; I'd not defend correcting much beyond it.

Issue #221, opened 2026-08-28
Geocoding: forward results drop parsed address detail, no reverse endpoint, provider unpublished


hokiepokedad2:
PoracleWeb.NET calls the configured geocoder **directly**, using `providerURL` from `/api/config/poracleWeb` and speaking Nominatim's query and response format. That works until an operator sets `provider = "photon"`, at which point we are pointed at a Photon instance still sending `/search?format=json&addressdetails=1` and the search box silently returns nothing.

The right fix on our side is obvious: stop calling providers and call `/geocode/forward`, which resolves through whichever provider is configured and hands back one shape, already cached and circuit-broken. Photon and Google then cost us nothing.

Three things stand in the way. The first is the blocker; the others are parity and cosmetics.

## 1. `ForwardResult` discards address detail it has already parsed

`geocoding.go:37`:

```go
type ForwardResult struct {
	Latitude  float64 `json:"latitude"`
	Longitude float64 `json:"longitude"`
	City      string  `json:"city,omitempty"`
	Country   string  `json:"country,omitempty"`
}
```

Four fields, against `Address`'s eighteen. And the detail is not unavailable — `nominatim.go:176` sets `addressdetails=1`, `:184` unmarshals the full address, and `:194` then keeps `city` and `country` and drops the rest on the floor.

What that costs a client, from a live instance:

```
GET /api/geocode/forward?q=1600 Pennsylvania Avenue Washington

[{"latitude":38.8792256,"longitude":-76.9819501,"city":"Washington","country":"United States"},
 {"latitude":38.8975196,"longitude":-77.0375588,"city":"Washington","country":"United States"},
 {"latitude":38.8976387,"longitude":-77.0365528,"city":"Washington","country":"United States"},
 {"latitude":38.8973756,"longitude":-77.0374079,"city":"Washington","country":"United States"},
 {"latitude":38.8974728,"longitude":-77.0365505,"city":"Washington","country":"United States"}, ...]
```

A location picker renders one row per candidate. These are five rows reading "Washington, United States", distinguishable only by coordinates the user cannot see. Against the provider directly the same query returns five distinct street addresses.

So the endpoint can answer "roughly where is this place" but not "which of these did you mean", and a picker is the main reason a client wants forward geocoding at all.

[...]

PR #224, opened 2026-08-31
Geocoding: keep parsed address detail on forward, add reverse endpoint (#221)

jfberry:
Addresses #221 items 1 and 2. Item 3 was already done.

## 1. `ForwardResult` keeps the detail it parses

It now carries `displayName`, `name`, `streetNumber`, `streetName`, `state` and `zipcode` alongside the existing `city` and `country` — all `omitempty`, so a field a provider does not supply is omitted rather than sent blank, and existing consumers are unaffected.

```
GET /api/geocode/forward?q=1600 Pennsylvania Avenue Washington

[{"latitude":38.8977,"longitude":-77.0365,
  "displayName":"White House, 1600, Pennsylvania Avenue Northwest, Washington, 20500, United States",
  "name":"White House","streetNumber":"1600","streetName":"Pennsylvania Avenue Northwest",
  "city":"Washington","state":"District of Columbia","zipcode":"20500","country":"United States"}, ...]
```

One thing the issue framed as a Nominatim problem is all three providers doing the same narrowing — Photon has the parts in its properties block, Google in its typed component list. So this is deleting a narrowing in three places, not adding parsing.

## 2. `GET /api/geocode/reverse?lat=&lon=&language=`

Returns the full `Address`. Coordinates are bounds-checked in the schema, so an impossible one is a 422 naming the field rather than a lookup that quietly resolves nothing. No address is **404**, not 500 — nothing failed, and that also covers `[geocoding] forward_only`, where reverse is switched off entirely. 503 when no geocoder is configured, matching forward.

## 3. `provider` was already published

It landed in #199 (TurtIeSocks) after 5.2.1, which is why it is absent from the build you tested. Its doc string did list `nominatim || photon` and omit `google` — corrected, and it now says explicitly not to infer configuration from `providerURL`, since google uses none.

## One correction to the issue's premise

The issue describes the forward endpoint as "already cached and circuit-broken". That is true of **reverse**, not forward: `Geocoder.Forward` calls the provider directly with no cache and no breaker.

Moving onto the endpoint still gets you provider-agnosticism and one response shape — the thing that actually breaks today when an operator switches to Photon — but budget for the same request volume reaching the provider, not less. Reverse is where the cache is, which is a further argument for adopting both halves rather than just search.

## Verification

`go build`, `go vet`, `go test ./...` (45 packages) and `golangci-lint` clean. Written test-first, with the failing test observed before each change. OpenAPI golden regenerated; `/geocode/reverse` appears in the spec.

### DIFF
--- a/processor/cmd/processor/main.go
+++ b/processor/cmd/processor/main.go
@@ -385,6 +385,7 @@ func main() {
 	api.RegisterStatsShiny(humaAPI, proc.stats.ExportShinyStats)
 	api.RegisterStatsShinyPossible(humaAPI, proc.stats.ExportShinyPossible)
 	api.RegisterGeocode(humaAPI, proc.enricher.Geocoder)
+	api.RegisterGeocodeReverse(humaAPI, proc.enricher.Geocoder)
 
 	// poracle-test (migrated to huma, in place — same path, same {"status":"ok"}
 	// body, open `webhook` request part, problem+json errors).
--- a/processor/internal/api/geocode_reverse_test.go
+++ b/processor/internal/api/geocode_reverse_test.go
@@ -1,0 +1,126 @@
+package api
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/gin-gonic/gin"
+
+	"github.com/pokemon/poracleng/processor/internal/geocoding"
+)
+
+// fakeReverseGeocoder records what it was asked for and returns a fixed address.
+type fakeReverseGeocoder struct {
+	addr     *geocoding.Address
+	gotLat   float64
+	gotLon   float64
+	gotLang  string
+	forwards int
+}
+
+func (f *fakeReverseGeocoder) Forward(string) ([]geocoding.ForwardResult, error) {
+	f.forwards++
+	return nil, nil
+}
+
+func (f *fakeReverseGeocoder) GetAddressForLanguage(lat, lon float64, language string) *geocoding.Address {
+	f.gotLat, f.gotLon, f.gotLang = lat, lon, language
+	return f.addr
+}
+
+func reverseAPI(t *testing.T, g ReverseGeocoder) *gin.Engine {
+	t.Helper()
+	gin.SetMode(gin.TestMode)
+	r := gin.New()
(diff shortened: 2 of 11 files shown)


jfberry:
Photon `displayName` fixed in `b06af853` on the PR branch — good catch, and it was the field I'd just finished calling the most useful one for a picker row.

Composed via the existing `formatOpenCage` rather than assembling one by hand. Photon's `Reverse` already builds a formatted address that way, using country-specific OpenCage templates with a comma-separated fallback, so forward and reverse now render the same address identically instead of two formats drifting apart:

```
GET /api/geocode/forward?q=1600 Pennsylvania Ave    (provider = photon)

{"latitude":38.8977,"longitude":-77.0365,
 "displayName":"Pennsylvania Avenue Northwest 1600, 20500 Washington, United States",
 "name":"White House","streetNumber":"1600","streetName":"Pennsylvania Avenue Northwest",
 "city":"Washington","state":"District of Columbia","zipcode":"20500","country":"United States"}
```

One trap that came with it, now covered by a regression test: `formatOpenCage` **mutates** the components map — it deletes `country` when the provider is configured `includeCountry=false`. Composing before reading `Country` would have silently blanked that field for those operators.

## On converging `ForwardResult` and `Address`

Deliberate, and we're going to leave it. The asymmetry is real but it isn't only field drift.

`Address` carries `addr` and `flag`. `addr` is rendered from a **per-operator Handlebars template** and `flag` is a country flag emoji — both exist to feed DTS alert templates, not to describe a place. A search hit has no meaningful value for either, so converging would mean either fabricating them or shipping a wider type with two permanently empty fields, which is worse for you rather than better.

The stronger reason is coupling. `Address` is the domain type the enrichment and alert pipeline consume. Making it the forward response too would mean a future DTS requirement lands as a breaking change to a public API. Keeping the API type separate from the alert-rendering type is the thing preventing that, and it's worth the flat-versus-nested remap on your side.

The one field where you have a point rather than a category difference is `countryCode` — same concept, and all three providers supply it. It isn't in this PR. If it turns out you want it for flag or locale formatting, open a line for it and it's a one-field additive change rather than a reshape.

`name` stays forward-only for the mirror-image reason: it's the place's own name when it has one distinct from its address, which is a search-result concept.

Merging once CI is green.

Issue #227, opened 2026-09-15
v2 pokemon: the pokemon_id=0 catch-all is readable but not writable, so GET then PUT fails


hokiepokedad2:
`V2PokemonRule` cannot express the catch-all rule, and reads one back anyway, so a `GET` then `PUT` of an untouched rule fails. Same shape as #209, on the most common tracking type.

## The rule exists and is first-class

`processor/internal/matching/pokemon.go:122`:

```go
// Basic Pokemon - everything (pokemon_id=0 catch-all)
matched = append(matched, m.matchMonsters(pokemon, st.Monsters.ByPokemonID[0], pokemon.PokemonID, pokemon.Form, true, 0, pvp.LeagueRank{})...)
```

It runs before the by-id branch and matches every spawn. The bot creates these through `!track everything`, and `everything_flag_permissions` exists specifically to govern who may. So this is a supported feature with its own permission mode, not a stray value.

## v2 declares it unwritable

`processor/internal/api/v2_pokemon.go:22`:

```go
PokemonID int `json:"pokemon_id" required:"true" minimum:"1" doc:"Pokédex id (required)"`
```

`required`, `minimum 1`, and a plain `int` rather than a pointer, so 0 cannot be sent and the field cannot be omitted either. There is no way to write the rule.

## But the read emits it

`v2_pokemon.go:169` copies the stored value through unchanged:

```go
PokemonID: row.PokemonID,
```

So `GET /v2/humans/{id}/tracking/pokemon` returns `"pokemon_id": 0` for a catch-all rule, and `PUT`ting that exact body back is a 422 on `minimum 1`. A client can read the rule, render it, and not save it.

That is the round-trip invariant from #209 — "a read emits exactly one targeting field, and it is whatever the rule is stored as" — broken in the other direction: the read emits a value the write refuses.

## Scale

653 rules on one instance, which is about 3% of its Pokemon rules. They are not dormant: the catch-all branch fires on every spawn, so these are among the most active rules on the server.

For PoracleWeb.NET they currently fall back to v1 on every edit, which works while v1 exists. It is frozen, so on the day it goes those rules become uneditable through the API entirely.

## The fix you already chose for this

[...]

Issue #229, opened 2026-09-16
v2 GET /humans: list projection omits last_checked, disabled_date and notes, so it cannot back an admin grid


hokiepokedad2:
Thank you for #214 — `GET /v2/humans` and `DELETE /v2/humans/{id}` on `develop` are exactly the shape asked for, and the comma-separated `id=a,b,c` matching the bulk tracking delete is a nice touch.

One gap stops the list replacing a direct database read on this end. Its projection is:

```
admin_disable, current_profile_no, enabled, id, language, name, type
```

The admin user list this would replace renders three more columns, all of which `GET /v2/humans/{id}` already returns for a single human:

- `last_checked` — when Poracle last saw them, the column admins sort by to find dormant accounts
- `disabled_date` — paired with `admin_disable`, so a disabled account shows *when*
- `notes` — the free-text admin note

Fetching them per-id is not an option at this scale: the instance this is for has 2333 humans, so the list is the only viable read and the per-human GET is 2333 calls.

Would you consider adding those three to the list item? They are already selected for the single GET, so it should be the same projection rather than new query work. If the concern is payload size for callers that do not want them, an opt-in `?include=notes,timestamps` or a `?verbose=true` would serve equally well — the important thing is that one call can answer the grid.

For context on why this matters here rather than being a nicety: these three columns are the only reason this application still maps Poracle's `humans` table in code. With them, the last admin read goes through the API and the direct database dependency can be deleted outright. Without them it stays for one query.

Two smaller notes from the same pass, neither of them blocking:

- `DELETE /v2/humans/{id}` cascading tracking rules and summary schedules is better than what this end does today, which deletes alarms in a separate step that can fail and leave rows orphaned. Worth calling out in release notes — it is a robustness improvement for anyone doing account deletion through the API.
- `?type=` filtering is what makes the webhook list work without a database read. Confirming that `type=webhook` is the exact stored value a webhook human carries would save the next person a guess.

jfberry:
PR up for review: #230. CI green, targets `develop`.

`last_checked`, `disabled_date` and `notes` added to the list item. You were right that it is the same projection — `humanRowColumns` already selects all three for the single GET, so there was no new query work.

Added unconditionally rather than behind `?include=` or `?verbose=`: three fields on a list that already returns seven, against a 2,333-row grid, does not justify making callers opt in to the common case. If payload turns out to matter at a larger scale, a filter is easier to add later than an unconditional field is to remove.

[...]

Issue #228, opened 2026-09-16
v2 setAreas: trusted also bypasses the community area restriction, which its own comment says it does not


hokiepokedad2:
`trusted: true` on `POST /v2/humans/{id}/areas` lifts the community area restriction as well as the `userSelectable` filter. The code comment beside the flag says it does not:

```go
// This lifts the userSelectable filter
// ONLY — an unknown fence name is still rejected, so a typo cannot become
// a stored area that matches nothing. See #215.
Trusted bool `json:"trusted,omitempty" ...`
```

But `settableAreaNames` folds both filters behind one flag (`processor/internal/api/v2_humans.go`, on `develop` at `634fe044`):

```go
unrestricted := admin || trusted

var allowed []string
for _, f := range st.Fences {
    if !unrestricted && !f.UserSelectable {
        continue
    }
    allowed = append(allowed, strings.ToLower(f.Name))
}
if deps.Config != nil && deps.Config.Area.Enabled && !unrestricted &&
    human != nil && human.AreaRestriction != nil {
    allowed = community.FilterAreas(
        deps.Config.Area.Communities, human.CommunityMembership, allowed)
}
```

So the `community.FilterAreas` call is skipped for a trusted request too.

## Reproduced on a build of `develop`

`[area_security] enabled = true`, human in one community whose `allowed_areas` are `Avoca Beach, Erina, Holgate, Kincumber, Terrigal`. `Aberdeen` belongs to a different community.

```
GET  /api/v2/humans/{id}/areas
  -> 5 areas: Avoca Beach, Erina, Holgate, Kincumber, Terrigal

POST /api/v2/humans/{id}/areas  {"areas":["Aberdeen","Erina"]}
  -> {"areas":["erina"],"rejected":["Aberdeen"]}                     correct

POST /api/v2/humans/{id}/areas  {"areas":["Aberdeen","Erina"],"trusted":true}
  -> {"areas":["aberdeen","erina"],"rejected":[]}                    restriction bypassed
```

## Why it matters for a client

The case `trusted` was added for (#215, thank you) is a server-side client asserting "the user drew this fence, so let them select it". That client is typically also the thing relaying the user's own requested area list. If the two travel in one call — and `setAreas` is a whole-list replace, so they must — then honouring the user's list under `trusted` hands them every area on the instance, area restriction included.

[...]

PR #230, opened 2026-09-16
v2: catch-all round-trip, trusted scope, admin-grid columns (#227, #228, #229)

jfberry:
Fixes #227, #228 and #229, plus the two related asks from #228.

Two of these were mine, and worth saying so plainly.

## #227 — a regression I introduced

`minimum:"1"` on `pokemon_id` came from the bounds work in `8fbe1e4d`. Before that the catch-all **was** writable: `required` only means the key is present, so `{"pokemon_id": 0}` was accepted and stored. I closed a working path and left the read emitting the value, which is what broke the round trip.

`pokemon_id` is now `*int` without `required` — omitted or null is the catch-all, stored as 0, read back as null. Blank is the wildcard, as on invasion.

The same bound bit `pvp_ranking_best`, `pvp_ranking_worst`, `rarity` and `size`. Those get the **opposite** treatment, because 0 there is not a valid rank or tier: the read reports the wildcard via `ptrUnlessAny`, so the rule round-trips and writing it back stores the canonical value. Loosening the bound instead would let clients write more of the junk.

**One correction to the issue's premise.** Those `pvp_ranking_best = 0` rows were not "written by PoracleNG itself" — no current path writes 0:

- the column defaults to `1` (`000001_initial_schema.up.sql:64`)
- the bot sets `Best: 1` explicitly for non-PVP rules, with a comment naming this exact hazard ("surfacing 0/0 instead of null in the v2 API")
- v1 and v2 both default to 1 when the field is absent

The one way a 0 gets in is a client sending `"pvp_ranking_best": 0` **explicitly** through v1, where `flexInt` passes it straight through. Worth checking against your own write path — otherwise the read is fixed here and new 0s keep arriving from there.

## #228 — also mine, and the comment was the tell

`settableAreaNames` folded both filters behind `unrestricted := admin || trusted` while the doc comment beside the flag said it lifted `userSelectable` **only**. Split, so `admin` lifts both and `trusted` lifts only `userSelectable`. Your suggested patch was the right one.

## #229

`last_checked`, `disabled_date` and `notes` added to the list item. All three were already selected by `humanRowColumns`, so it is the same projection. Unconditional rather than behind `?include=` — three fields on a list that already returns seven does not warrant an opt-in.

To confirm your second note: **`webhook` is the exact stored value**. `?type=webhook` is what you want.

## The two asks

**`?profile=` and `?all_profiles=` on set-areas.** It wrote the active profile only. `?profile=` mirrors the tracking endpoints and set-location; `?all_profiles=true` is the shape that actually answers "the user deleted a fence, drop it everywhere" in one call. Mutually exclusive, 422 together.

**`?trusted=` on v2 tracking writes** — the per-rule counterpart of the set-areas body flag, so an alarm can be confined to a fence the user drew. Same scope as #228 leaves it: `userSelectable` only, unknown names still rejected, community restriction still applied. v1 call sites pass `false`; it is frozen and never trusted.

## Also fixed while in there

`MockHumanStore.SetArea` ignored `profileNo` entirely and wrote only `humans.area`, so a test could not tell a per-profile write from a global one. It now mirrors `SQLHumanStore`. Second time a mock has diverged from production in a way that would let a wrong test pass — the first was `AddProfile` in #217.

## Verification

`go build`, `go vet`, `go test ./...` (45 packages) and `golangci-lint` clean. Written test-first with the failing test observed before each change. Four existing tests encoded contracts this deliberately reverses and were updated rather than worked around; OpenAPI golden regenerated.


hokiepokedad2:
Built `fix/v2-api-round-trip-and-areas` (`dae652da`) and ran it against a real instance. **The mixed case does not work**, and it is the one you flagged.

## Reproduction

Human in one community, `allowed_areas` = `Avoca Beach, Erina, Holgate, Kincumber, Terrigal`. `Aberdeen` is an admin fence in a different community. `Animal Kingdom` is served `userSelectable=false` — the stand-in for a user-drawn fence.

With `[area_security] enabled = true`:

```
POST /api/v2/humans/{id}/areas  {"areas":["Erina","Aberdeen","Animal Kingdom"]}
  -> {"areas":["erina"],"rejected":["Aberdeen","Animal Kingdom"]}

POST /api/v2/humans/{id}/areas  {"areas":["Erina","Aberdeen","Animal Kingdom"],"trusted":true}
  -> {"areas":["erina"],"rejected":["Aberdeen","Animal Kingdom"]}     <-- identical
```

`Aberdeen` being rejected is correct and is exactly what #228 asked for. But `Animal Kingdom` is rejected too, and that is the fence `trusted` exists to allow.

With `[area_security] enabled = false`, the same request is fine:

```
  -> {"areas":["erina","aberdeen","animal kingdom"],"rejected":[]}
```

So the two filters are no longer folded together — that part is fixed — but they now compose in a way that cancels the lift.

## Cause

A user-drawn fence appears in **no** community's `allowed_areas`. `trusted` adds it to `allowed`, and `community.FilterAreas` then removes it again:

```go
for _, f := range st.Fences {
    if !admin && !trusted && !f.UserSelectable {
        continue
    }
    allowed = append(allowed, strings.ToLower(f.Name))   // Animal Kingdom lands here
}
if deps.Config != nil && deps.Config.Area.Enabled && !admin && ... {
    allowed = community.FilterAreas(...)                 // and is stripped here
}
```

Under `area_security`, `trusted` is therefore a no-op for precisely the fences it is for.

## A possible shape

Exempt the fences the lift added from the community filter, rather than running the filter over the merged list:

```go
var allowed, lifted []string
for _, f := range st.Fences {
    name := strings.ToLower(f.Name)
    if !f.UserSelectable {
        if admin || trusted {
            lifted = append(lifted, name)
        }
        continue
    }
    allowed = append(allowed, name)
}
if deps.Config != nil && deps.Config.Area.Enabled && !admin &&
    human != nil && human.AreaRestriction != nil {
    allowed = community.FilterAreas(
        deps.Config.Area.Communities, human.CommunityMembership, allowed)
}
allowed = append(allowed, lifted...)
```

The argument for it: `area_security` governs which **published** areas a user may subscribe to, and a fence the user drew is not a published area — it is in no community by construction, so filtering it by community can only ever reject it.

The argument against, which is yours to weigh: `trusted` lifts `userSelectable` for *every* non-selectable fence, not only ones this user drew, because nothing in the request says whose fence it is. So this would let a secret-holding caller put a user on any non-selectable fence regardless of community. That is much narrower than the original bug — non-selectable fences only, and only for a caller already asserting authorisation — but it is not nothing, and if you would rather `trusted` simply not serve this case under `area_security`, that is a legitimate answer. It would mean clients running `area_security` keep a direct write for user fences, which is where we are today, and worth saying so in the docs.

A third option, if you want the tighter version: have the caller name the fences it is vouching for (`"trusted_areas": ["animal kingdom"]` alongside `areas`), so the assertion is per-fence rather than per-request. More work, and it makes the flag's scope explicit rather than implied.

## The other two halves check out

**#229** — `last_checked`, `disabled_date` and `notes` are all present on the list item and the null-not-zero-date handling is right. Checked against the actual grid: `disabled_date` and `notes` are the two it renders. `last_checked` is on our model but no column reads it today — still worth having, just not load-bearing, so do not let it hold the PR up. No need for `fails`. Confirming `type=webhook` saved a guess, thank you.

**#227** — your read is right and the 0s are ours. `MonsterCreate.PvpRankingBest` has no initializer where its sibling has `PvpRankingWorst = 4096`, so it goes out as 0 through v1's `flexInt`. Production here: **15,383 rows at 0** against 7,770 at 1. Fixing on our side so the repair is not undone by the next write; thank you for spotting that the fix alone would not have held.

PR #232, opened 2026-09-17
fix(delivery): optionally drop sends on full destination lanes

lucabravi:
## Problem

A single rate-limited destination can fill its per-destination delivery lane. Enqueuing another send currently blocks a shared render worker.

If enough render jobs target that destination, all render workers can become blocked and the global render queue can fill, delaying unrelated destinations.

## Change

Add the optional tuning setting:

```toml
delivery_queue_drop_on_full = false
```

When enabled, a new send is dropped if its destination lane is already full, allowing the render worker to continue processing other destinations.

The default remains `false`, preserving the existing blocking behavior. The default `delivery_queue_size` also remains unchanged at `200`.

The existing backpressure metric is retained and now counts every enqueue attempt that encounters a full destination lane, regardless of the selected policy.

## Tests

- Verify that a full lane drops new sends when the option is enabled.
- Verify that the default behavior still blocks until capacity becomes available.
- `go test ./... -count=1`


jfberry:
Thanks for this — the report and the diagnosis were both spot on. You correctly identified that the blocked `Dispatch()` calls were occupying the shared render workers, which is exactly the mechanism, and that's the part that was hard to find.

I've picked up the fix in #235 rather than send you round another review cycle, because digging into it turned up a few things that changed the shape of the remedy:

**It isn't only a "single rate-limited destination" problem.** `processJob` calls `WaitForRateLimit` synchronously in the lane drainer, and on Discord that spins on a global 50 req/sec token bucket shared by *every* Discord destination. Above ~50/s system-wide, every lane drains slowly at once — so drop-on-full would have shed across all destinations rather than quarantining the noisy one. Your Telegram case was one trigger for a more general structural issue.

**Dropping the arriving job sheds the wrong end.** A lane only fills because its drainer is parked in backoff, so the 200 buffered jobs are the stalest and the arrival is the freshest. `processJob`'s expiry gate would have thrown most of those buffered ones away at the head anyway. #235 sheds the oldest instead, so the buffer stays useful.

**Bypass jobs needed a carve-out.** This one was subtle and would have bitten hard in production: rate-limit breach notices and ban farewells go through `DispatchBypass` precisely *because* the destination is flooded — so their lane is full by definition. Under the flag they'd be dropped, and a user could be auto-disabled after `max_limits_before_stop` breaches having never received a single warning.

**Made it unconditional rather than a flag.** Stalling every render worker behind one destination isn't really a behaviour anyone would opt into, and default-`false` would have left the bug you hit unfixed for everyone who hadn't yet diagnosed it. There's precedent in the repo too — `899d364a` and `c461c906` fixed the same bug class structurally when a backing-off request pinned the platform semaphore. The render worker was just the next shared resource up the stack.

Plus a dedicated `poracle_delivery_lane_shed_total` and a `shed_lane_full` result label so the loss is actually alertable, which your PR raised the right question about.

You're credited as `Co-Authored-By` on the commit. Closing this in favour of #235 — please do give it a look, and thank you for reporting it properly instead of just raising `delivery_queue_size` and moving on.

Issue #236, opened 2026-09-19
Telegram: no global send-rate cap, no proactive pacing — decide whether to build it


jfberry:
Follow-up from #232 / #235. Raising this to be *decided*, not assumed — after digging into it I think one part is worth doing, one part is marginal, and one part doesn't do what it looks like it does. Benefits spelled out below with the case against each.

## Where we are today

`TelegramSender.WaitForRateLimit` is a no-op ([telegram.go:140](https://github.com/jfberry/PoracleNG/blob/develop/processor/internal/delivery/telegram.go#L140)) — Telegram has **no proactive rate limiting of any kind**. Limits are discovered by taking a 429 and sleeping.

That reactive handling is solid and not in question: both 429 branches read `parameters.retry_after`, default to 1s when the (optional) `parameters` object is absent, cap absurd values at 60s rather than sleeping on them, and back off cancellably. `currentBackoffUntil` is written but only read by the status snapshot — it gates nothing.

Telegram publishes no proactive quota signal. There is no equivalent of Discord's `X-RateLimit-Limit/Remaining/Reset/Bucket`, which is exactly what lets `DiscordRateLimiter` stay ahead of the limit. A successful `sendMessage` tells us nothing about remaining budget. So anything proactive here must be built from published constants plus inference from observed 429s — it cannot be a port of the Discord limiter.

Published limits ([Bot FAQ](https://core.telegram.org/bots/faq)): ~1 msg/sec to a single chat, **20 msgs/minute to a group**, ~30 msgs/sec bulk across different chats. The FAQ explicitly adds: *"We may allow short bursts that go over this limit, but eventually you'll begin receiving 429 errors."*

---

## Part A — global 30/sec cap (the part I think is real)

**We have no aggregate Telegram send-rate limit at all.** Discord has a global 50 req/sec token bucket (`ratelimit.go:66`) that every destination passes through. Telegram has nothing equivalent.

**Benefit:** above ~30 sends/sec instance-wide, *every* Telegram destination starts 429ing at once, regardless of how well-behaved each individual chat is. That's a correlated, instance-wide degradation with no single culprit to point at — and it's the precise failure mode the Discord global bucket exists to prevent. A busy instance with many Telegram channels reaches 30/sec long before any individual group reaches 20/min.

**Cost:** low. `tokenBucket` in `ratelimit.go` is already written, tested, and platform-agnostic; this is wiring it into the currently-empty `WaitForRateLimit`, which `processJob` already calls per job.

[...]

PR #237, opened 2026-09-20
feat(telegram): add proactive global send rate limiting

lucabravi:
## Summary

Implements Part A of #236 by adding a proactive global Telegram message-send limiter shared by delivery and the polling/reconciliation bot when they use the same bot token.

- defaults to a sustained rate of 29 message sends per second with a burst capacity of 5
- counts individual outbound message API calls, including every part of a multi-message notification and each retry
- covers `sendMessage`, `sendPhoto`, `sendSticker`, `sendLocation`, `sendVenue`, and `sendDocument`
- does not count edits, deletes, polling, or read-only Bot API calls
- waits before acquiring the Telegram HTTP semaphore and remains context-cancellable
- keeps the existing reactive 429 / `retry_after` handling as a second line of defence

The configured rate is a sustained token-bucket rate, not a strict rolling-window ceiling. The default burst of 5 intentionally preserves the short burst tolerance documented by Telegram.

## Configuration

Adds two `[tuning]` settings:

```toml
telegram_global_rate_per_second = 29
telegram_global_burst = 5
```

Both values must be greater than zero when explicitly configured. Existing installations receive the defaults without requiring config changes.

## Observability

Limiter wait time is reported through the existing `poracle_delivery_rate_limit_wait_seconds` metric with `platform="telegram"`. The Grafana panel is renamed from Discord-specific to delivery-wide wording.

## Scope

This PR deliberately does not add static per-chat pacing or paid broadcasts. Those are separate decisions in #236, and the global limiter does not claim to solve an overloaded single-group workload.

## Verification

- `gofmt`
- `go build ./...`
- `go test ./...`
- `go test -race -count=1 ./...`
- `go vet ./...`
- `golangci-lint run ./...` (`0 issues`)


jfberry:
I'm not a fan of magic settings that users can override without any understanding. I'd prefer to hard code these values if they are correct. How did you come to them? 
I'll get claude to do a full review

lucabravi:
> I'm not a fan of magic settings that users can override without any understanding. I'd prefer to hard code these values if they are correct. How did you come to them? I'll get claude to do a full review

Same here, but there is no absolute 'correct value' in this case since Telegram applies dynamic limits from bot to bot. Based on the test with my own bot, I have conservatively hardcoded the baseline to 29 messages per second (just under the official 30 msgs/s threshold) and allowed a burst of 5. While these numbers proved to be stable during my tests, they aren't universally correct or set in stone. Telegram's anti-spam system is opaque and heavily scales limits based on bot age, reputation, and user engagement... creating a dynamic limit seems excessive to me; I preferred a conservative approach.

Anyway it is only a proposal :)

